PR #202, opened 2024-02-22
Fix cursor mutiline

phong9xNcc:


libs/components/src/lib/components/MessageWithUser/index.tsx
```diff
@@ -39,20 +39,29 @@ function MessageWithUser({ message, preMessage }: MessageWithUserProps) {
 	}, [timeDiff, message, preMessage]);
 
 	const renderMultilineContent = () => {
-		const lines = content.replace(/<br>/g, '\n').split('<br>');
+		const lines = content.split('\n');
+		const mentionRegex = /(@\S+?)\s/g;
```
nguyentrannhan, line 43: var re = /\B@\w+/g; 
var str = 'The @dog went to the park.\nBut not here: The d@og went to the park.\nOr here: The@dog went to the park.';
var res = str.match(re);
document.body.innerHTML = "<pre>" + JSON.stringify(res, 0, 4) + "</pre>";

PR #270, opened 2024-02-27
Feat emoji suggestion

phong9xNcc:


libs/components/src/lib/components/MessageBox/index.tsx
```diff
@@ -201,6 +320,7 @@ function MessageBox(props: MessageBoxProps): ReactElement {
 						plugins={plugins}
 						ref={editorRef}
 						handlePastedFiles={onPastedFiles}
+						// handleBeforeInput={handleBeforeInput}
```
nguyentrannhan: delete if not used

Issue #421, opened 2024-03-15
Make a channel name bold when has unread message


nguyentrannhan:
![Image](https://github.com/nccasia/mezon-fe/assets/17939754/4b1d2416-4e9b-4d2e-840b-f4790431987b)

VyPham23:
- Add red number on the right for mentioning

PR #458, opened 2024-03-20
 work on task  #421, fix bug UI setting clan.

thuthuy1901:
work on task  #421, fix bug UI setting clan.


libs/store/src/lib/channels/channels.slice.ts
```diff
@@ -97,6 +97,9 @@ export const fetchChannels = createAsyncThunk('channels/fetchChannels', async ({
 	}
 
 	const channels = response.channeldesc.map(mapChannelToEntity);
+	thunkAPI.dispatch(channelsActions.setArrayUnreadChannel(channels));
+
+	console.log(channels);
```
nguyentrannhan: remove console.log

PR #541, opened 2024-03-28
Panel UI channel

anbuihoang:
Panel UI channel


VyPham23:
- Change panel background to color `#535353` or `#6A6A6A`
- Change hover area color to color `#0040C1` or `#004EEB`

PR #551, opened 2024-03-29
fix jump to message

Cut666:


libs/components/src/lib/components/NotificationList/NotifyMentionItem.tsx
```diff
@@ -104,6 +111,7 @@ function NotifyMentionItem({ notify }: NotifyMentionProps) {
 					attachments={data.attachments}
 					mentions={data.mentions}
 					mode={2}
```
nguyentrannhan: hard code mode

thuthuy1901 (commented):
ok

VyPham23:
- `Jump` button feature works properly as expected
- Can jump to message on another channel ✅
- Can jump to message on the same channel ✅

Issue #550, opened 2024-03-29
Create Category modal is not sync with system modal UI


VyPham23:
**Actual:**

![Image](https://github.com/nccasia/mezon-fe/assets/90541812/67a7a02e-ce5e-4fe6-a6bf-8cc70a4b3e20)

**Expected:**
- Change modal background color to the same color as Create Channel background color
- Modify text and input field as Create Channel modal

![Image](https://github.com/nccasia/mezon-fe/assets/90541812/2ec29eab-48e5-4017-9076-2a4d3140d248)

Abilene-may:
### Verified

![Image](https://github.com/nccasia/mezon-fe/assets/107451757/977a62ac-c64d-47a0-ae43-d9a012742f60)

Issue #536, opened 2024-03-28
The logic of unread message does not work right


VyPham23:
- Check the logic of unread message

https://github.com/nccasia/mezon-fe/assets/90541812/d0f45f10-613e-4bf6-ad83-990b4d5e6a74

Abilene-may:
Verified the issue and confirmed that there are no more bugs.

Issue #512, opened 2024-03-26
control panel when joining to voice channel


nguyentrannhan:
![Image](https://github.com/nccasia/mezon-fe/assets/17939754/f94ea46c-ee50-444b-9be2-06e4d1e67152)

thuthuy1901:
add event for button

PR #554, opened 2024-03-29
task #512, #536, #550

thuthuy1901:


Cut666:
done

anbuihoang:
Done

VyPham23:
- Voice call panel UI is updated
- The logic of unread message is updated
- Create category modal UI is sync

PR #555, opened 2024-03-29
fix task 540 and fix icon voice control pannel

Thanh97x:
fix task 540 and fix icon voice control pannel


VyPham23:
- Fixed white area in right down corner of channel list

Issue #545, opened 2024-03-28
Text before mentioned username disappears when sending


VyPham23:
- Input:

![Image](https://github.com/nccasia/mezon-fe/assets/90541812/1e1c1346-7ff5-48d0-872d-ee6f6f42e226)

- Output:

![Image](https://github.com/nccasia/mezon-fe/assets/90541812/7efeb7c7-31ef-4eb9-917d-389876857248)

Abilene-may:
### Verified and confirmed there are no more bugs.
**input**


![Image](https://github.com/nccasia/mezon-fe/assets/107451757/49f05b5b-00fd-43ee-b0f0-cef0e511c5a8)

**output**


![Image](https://github.com/nccasia/mezon-fe/assets/107451757/6397a223-b5d7-4bc7-af71-22c7a090956b)

Issue #513, opened 2024-03-26
Create voice UI when use join to voice channel


nguyentrannhan:
![Image](https://github.com/nccasia/mezon-fe/assets/17939754/d5075c9e-b279-4a4e-96e2-b7281e48874f)

![Image](https://github.com/nccasia/mezon-fe/assets/17939754/cc12d255-c624-4442-8eb2-b130f122ae4f)

thuthuy1901:
![Image](https://github.com/nccasia/mezon-fe/assets/130724306/71d8cdaf-1238-47ec-a670-334a4afa4882)

PR #586, opened 2024-04-05
task #513

thuthuy1901:
task #513: Create voice UI when use join to voice channel, #545: Text before mentioned username disappears when sending.


Cut666:
done

PR #588, opened 2024-04-05
refactor message reply 0404

phong9xNcc:


libs/components/src/lib/components/MessageWithUser/MessageReply.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { IMessageWithUser } from '@mezon/utils';
+import { useEffect, useState } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import * as Icons from '../Icons/index';
+
+type MessageReplyProps = {
+	message: IMessageWithUser;
+};
+
+// TODO: refactor component for message lines
```
nguyentrannhan, line 11: Do we need to pay attention to this TODO?

PR #597, opened 2024-04-08
Feat 3button 0704

phong9xNcc:


libs/utils/src/lib/types/index.ts
```diff
@@ -326,3 +325,14 @@ export type SenderInfoOptionals = {
 	sender_name?: string | undefined;
 	avatar?: string | undefined;
 };
+
+export type IGif = {
```
nguyentrannhan: I prefix stand for Interface

Cut666:
done

Issue #598, opened 2024-04-08
There is no highlight when hover to message and the reply, emoji popup seems too high


nguyentrannhan:
![Image](https://github.com/nccasia/mezon-fe/assets/17939754/4d0dbe9d-0208-4cf9-acb9-bcb6ff4e4d74)

![Image](https://github.com/nccasia/mezon-fe/assets/17939754/7b6bcc4d-684d-47bd-8141-7e7511bc3898)

![Image](https://github.com/nccasia/mezon-fe/assets/17939754/be136150-7e9a-4a77-93bf-b12aad69918e)

PR #620, opened 2024-04-11
task #598, task #616, fix bug UI

thuthuy1901:
task #598: There is no highlight when hover to message and the reply, emoji popup seems too high.
task #616: add threedot menu. This property popup will appear when click to threedot button


Cut666:
done

PR #718, opened 2024-04-23
realtime delete message

anbuihoang:
realtime delete message


libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -311,6 +311,8 @@ export const messagesSlice = createSlice({
 					id: action.payload.id,
 					changes: action.payload,
 				});
+			} else if (action.payload.code === 2) {
```
nguyentrannhan: better use switch case in this case

anbuihoang: Done

PR #728, opened 2024-04-24
Fix 2000 char to file

phong9xNcc:


libs/components/src/lib/components/MessageBox/index.tsx
```diff
@@ -27,7 +27,7 @@ function MessageBox(props: MessageBoxProps): ReactElement {
 	const { attachmentDataRef, setAttachmentData } = useReference();
 
 	const onConvertToFiles = useCallback((content: string) => {
-		if (content.length > 2000) {
+		if (content.length > 4096) {
```
nguyentrannhan, line 30: should define CONSTANT

PR #741, opened 2024-04-25
anonymous message and click cannel, save edit message

Cut666:


libs/components/src/lib/components/MessageWithUser/MessageHead.tsx
```diff
@@ -50,7 +50,7 @@ const MessageHead = ({ user, message, isCombine }: IMessageHeadProps) => {
 					ref={panelRef}
 					onMouseDown={(event) => handleMouseClick(event)}
 				>
-					{username}
+					{username? username: 'Anonymous Message'}
```
nguyentrannhan: Change to just Anonymous

Cut666: done

PR #866, opened 2024-05-09
profileScreen

SonBeoDeveloper:
Screenshot 2024-05-09 174753

Screenshot 2024-05-09 174802


apps/mobile/src/app/screens/profile/styles.ts
phuocnguyenncc, line 1: Check lại width, height, fontSize, padding, margin k hardcode number, dùng verticalScale or horizontalScale

PR #883, opened 2024-05-13
fix: send attachment without text in message box

Tung2403:


libs/components/src/lib/components/MessageBox/ReactionMentionInput/index.tsx
```diff
@@ -166,7 +166,7 @@ function MentionReactInput(props: MentionReactInputProps): ReactElement {
 
 	const handleSend = useCallback(
 		(anonymousMessage?: boolean) => {
-			if (!valueTextInput || valueTextInput.trim() === '') {
+			if (!valueTextInput && attachmentDataRef.length === 0 || valueTextInput.trim() === '' && attachmentDataRef.length === 0) {
```
nguyentrannhan: add ((!valueTextInput && attachmentDataRef.length === 0) || (valueTextInput.trim() === '' && attachmentDataRef.length === 0))

PR #886, opened 2024-05-13
feat: handle send message text

phuocnguyenncc:


phuocnguyenncc:
https://github.com/nccasia/mezon-fe/assets/167954084/3807054e-43df-4499-99d1-55ea2c5aa016

PR #1012, opened 2024-05-22
Feat attachment file (docs, pdf, xlx, ...)  mobile

phuocnguyenncc:


phuocnguyenncc:
https://github.com/nccasia/mezon-fe/assets/167954084/9abe05bc-6ffe-48a3-8cff-0f727c2fbee0

PR #1021, opened 2024-05-22
create, delete channel is not notify to all clan.

thuthuy1901:
create, delete channel is not notify to all clan.


libs/store/src/lib/channels/channels.slice.ts
```diff
@@ -153,8 +164,7 @@ export const fetchChannelsCached = memoize(
 
 export const fetchChannels = createAsyncThunk('channels/fetchChannels', async ({ clanId, channelType = 1 }: fetchChannelsArgs, thunkAPI) => {
 	const mezon = await ensureSession(getMezonCtx(thunkAPI));
-	const response = await fetchChannelsCached(mezon, 100, 1, clanId, channelType);
-
+	const response = await mezon.client.listChannelDescs(mezon.session, 100, 1, '', clanId, channelType);
```
nguyentrannhan, line 167: chỗ này sao xóa cached đi em?

PR #1221, opened 2024-06-05
bug: mention channel not working on mobile and Interface error in notification screen

loi-huynhphuc:


apps/mobile/src/app/components/TextContent/index.tsx
phuocnguyenncc, line 1: Xóa file này, file này dup ra như k dùng

phuocnguyenncc (approved):

PR #1665, opened 2024-07-02
short user Anonymous, panel right click member list channel + DM + DM group, add status friend add, remove, pending

thuy1901:


libs/components/src/lib/components/DmList/DMListItem/index.tsx
```diff
@@ -12,6 +12,7 @@ export type DirectMessProp = {
 export type directMessageValueProps = {
 	type: string;
 	userId: string[];
+	DmID: string;
```
anbuihoang: Đổi tên biến theo quy tắc Camel Case

PR #1831, opened 2024-07-09
Fix chatbox limit height & remove icon created thread on DM

sadohpv:


apps/chat/src/styles.scss
```diff
@@ -551,7 +551,7 @@ input:-webkit-autofill:active {
 
 // css editorReactMention text box chat
 #editorReactMention {
-	max-width: calc(100% - 100px);
+	// max-width: calc(100% - 100px);
```
anbuihoang, line 554: Remove comment

libs/components/src/lib/components/Message/ChannelMessageOpt.tsx
```diff
@@ -1,154 +1,160 @@
...
+  const optionMenu = useOptionMenuBuilder(handleContextMenu);
+
+
+  const items = useMenuBuilder([reactMenu, replyMenu, editMenu, threadMenu, optionMenu]);
+
+  return (
+    <div className={`chooseForText z-[1] absolute h-8 p-0.5 rounded block -top-4 right-6 w-fit`}>
+      <div className="flex justify-between dark:bg-bgPrimary bg-bgLightMode border border-bgSecondary rounded">
+        <div className="w-fit h-full flex justify-between" ref={refOpt}>
+          {items.map((item, index) => {
```
anbuihoang, line 34: Using “if” unnecessarily

PR #1971, opened 2024-07-15
Fix reaction bottom

phong9xNcc:


apps/desktop/src/app/app.ts
```diff
@@ -73,7 +73,7 @@ export default class App {
 		App.mainWindow.setMinimumSize(950, 500);
 		App.mainWindow.setMenu(null);
 		App.mainWindow.center();
-
+		App.mainWindow.webContents.openDevTools();
```
nguyentrannhan: sao add devtool vao lam gi chu?

PR #2054, opened 2024-07-17
update show noti on mention tab

phong9xNcc:


libs/components/src/lib/components/NotificationList/index.tsx
```diff
@@ -26,8 +26,9 @@ function NotificationList({ unReadList }: NotificationProps) {
 		setCurrentTabNotify(valueTab);
 	};
 
-	const notificationItem = notification.filter((item) => item.code !== -9 && item.code === -10);
+	const notificationItem = notification.filter((item) => item.code !== -9 && item.code !== -10);
```
nguyentrannhan: Tao enum

phong9xNcc: e add r ah.

PR #2061, opened 2024-07-18
change component img for DM + short user, update display name for Ctrl+k, fix logic show kick member panel right+click

thuy1901:
change component img for DM + short user, update display name for Ctrl+k, fix logic show kick member panel right+click


libs/utils/src/lib/types/index.ts
```diff
@@ -565,3 +566,5 @@ export enum NotificationCode {
 	USER_REACTIONED = -10,
 	USER_REPLIED = -11,
 }
+
+export const IdUserAnonymous = '1767478432163172999';
```
nguyentrannhan: search .env

PR #2091, opened 2024-07-18
The conversation is auto scrolled and the receiver's info is hidden

anbuihoang:
The conversation is auto scrolled and the receiver's info is hidden


libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -176,13 +177,17 @@ export const fetchMessages = createAsyncThunk(
...
 		thunkAPI.dispatch(reactionActions.updateBulkMessageReactions({ messages }));
-		thunkAPI.dispatch(messagesActions.setQuatitiesMessageRemain(messages.length));
 
 		const hasMore = Number(response.messages.length) >= LIMIT_MESSAGE;
 		if (messages.length > 0) {
 			thunkAPI.dispatch(
 				messagesActions.setMessageParams({ channelId, param: { lastLoadMessageId: messages[messages.length - 1].id, hasMore } }),
 			);
+			messages.forEach((message) => {
+				if (message.code === EMessageCode.FIRST_MESSAGE) {
```
nguyentrannhan: Sao ko dispatch ở chỗ map luôn?

PR #2137, opened 2024-07-22
add permission admin for pin mess, kick member, fix continuous fetch api, name ava text

thuy1901:
add permission admin for pin mess, kick member, fix continuous fetch api, name ava text


libs/components/src/lib/components/ChannelSetting/Component/OverviewChannel/index.tsx
```diff
@@ -71,6 +71,7 @@ const OverviewChannel = (props: OverviewChannelProps) => {
 					value={channelLabel}
 					onChange={handleDisplayChannelLabel}
 					className="dark:bg-black bg-white pl-3 py-2 w-full border-0 outline-none rounded"
+					maxLength={64}
```
nguyentrannhan: define const MAX_LENGTH_CHANNEL_ALLOWED

PR #2195, opened 2024-07-24
fix bug role setting and display name DM

phatphanvan:


apps/mobile/src/app/screens/home/homedrawer/MessageItem.tsx
```diff
@@ -92,6 +93,8 @@ const MessageItem = React.memo((props: MessageItemProps) => {
 	const { markMessageAsSeen } = useSeenMessagePool();
 	const userProfile = useSelector(selectAllAccount);
 	const clanProfile = useSelector(selectUserClanProfileByClanID(currentClanId as string, message?.user?.id as string));
+	const user = useSelector(selectMemberByUserId(message.sender_id));
+
```
phuocnguyenncc: remove: get user from message

phuocnguyenncc (approved):

PR #2239, opened 2024-07-26
send link invite modal, ava anonymous, UX/UI role

thuy1901:
send link invite modal, ava anonymous, UX/UI role


libs/components/src/lib/components/ClanSettings/SettingRoleManagement/SettingListRole/index.tsx
```diff
@@ -19,17 +19,18 @@ type closeEditRole = {
...
+	const dispatch = useDispatch();
 	const appearanceTheme = useSelector(selectTheme);
 	const isChange = useSelector(getIsShow);
 
 	const clickRole = useSelector(getSelectedRoleId);
 	const [clickedRole, setClickedRole] = useState<null | string>(clickRole);
 	const nameRoleNew = useSelector(getNewNameRole);
 
-	const dispatch = useDispatch();
+	const isNewRole = useMemo(() => clickedRole === 'New Role', [clickedRole]);
```
sontung-nccvn: Primitive type values ​​that don't require complex calculations is no need to be wrapped in useMemo

libs/components/src/lib/components/ClanSettings/SettingRoleManagement/SettingListRole/index.tsx
```diff
@@ -55,31 +66,41 @@ const SettingListRole = (props: closeEditRole) => {
...
+type ItemRoleProps = {
+	title: string;
+	isChoose?: boolean;
+	onHandle?: () => void;
+}
+
+const ItemRole = forwardRef<HTMLDivElement, ItemRoleProps>((props, ref) => {
+	const {title, isChoose, onHandle} = props;
+	return (
+		<div ref={ref} onClick={onHandle ? () => onHandle() : undefined}>
```
sontung-nccvn: set default value for onClick instead of using Ternary operator

libs/components/src/lib/components/ListMemberInvite/ListMemberInviteItem.tsx
```diff
@@ -37,32 +28,58 @@ const ListMemberInviteItem = (props: ItemPorp) => {
...
+		/>
+	) : (
+		<ItemInviteUser 
+			userId={user?.id || ''}
+			avatar={user?.user?.avatar_url || ''}
+			displayName={user?.user?.display_name || ''}
+			userName={user?.user?.username || ''}
+			isInviteSent={isInviteSent ?? true}
+			onHandle={() => handleButtonClick('', 0, user?.id)}
+		/>
```
sontung-nccvn, line 62: consider add default props for these components instead of using ```||``` operator so often

PR #2488, opened 2024-08-07
fix: new red line, hashtag in dm

phatphanvan:


apps/mobile/src/app/screens/home/homedrawer/components/ForwardMessage/index.tsx
```diff
@@ -75,6 +75,9 @@ const ForwardMessageModal = ({ show, message, onClose }: ForwardMessageModalProp
 	}, [dmGroupChatList, listChannels])
 
 	const filteredForwardObjects = useMemo(() => {
+		if (searchText.trim().charAt(0) === '#') {
```
phuocnguyenncc: check valid data searchText before trim or add ?.

phuocnguyenncc (approved):

Issue #2448, opened 2024-08-06
Mobile App: DM/Channel msg > The uploaded attachment(s) can be sent multiple times when keep tapping the Send button


trucnguyen31:
## Steps:

1. Attach images/videos/files to the upload section
2. Keep tapping the Send button multiple times
3. Observe

## ER: 
The attachment(s) should be removed from the upload section after being sent once

## AR:
- The attachment is still kept in the upload section
- The uploaded attachment(s) can be sent multiple times when keep tapping the Send button

https://github.com/user-attachments/assets/4dc4f829-2d46-4912-b64a-7c88bb080240

trucnguyen31:
done

https://github.com/user-attachments/assets/0fa1f0b4-736b-4a7a-9550-653fbefdb821

PR #2599, opened 2024-08-12
Mobile fix spam send message

codengoo:
Mobile App: DM/Channel msg > The uploaded attachment(s) can be sent multiple times when keep tapping the Send button
[#2448](https://github.com/nccasia/mezon-fe/issues/2448)


phuocnguyenncc:
@codengoo này anh fix trc đó r nhé

PR #2617, opened 2024-08-13
Fix reply contain image

phong9xNcc:


libs/components/src/lib/components/MarkdownFormatText/PlainText.tsx
```diff
@@ -5,21 +5,17 @@ import React, { useMemo } from 'react';
 type PlainTextOpt = {
 	text: string;
 	isSearchMessage?: boolean;
-	isSingleLine: boolean;
-	isJumMessageEnabled: boolean;
-	isHover: boolean;
+	// isSingleLine: boolean;
+	// isJumMessageEnabled: boolean;
+	// isHover: boolean;
```
Tungnqs: delete comment di azai

Tungnqs (approved):

PR #2675, opened 2024-08-15
Fix info in reply reaction

phong9xNcc:


libs/core/src/lib/chat/hooks/useChatSending.ts
```diff
@@ -60,14 +74,50 @@ export function useChatSending({ channelId, mode, directMessageId }: UseChatSend
...
+	useEffect(() => {
+		if (!idNewMessageResponse || filteredLinkResults.length === 0) return;
+		const result = checkContentContainsOnlyUrls(filteredContent.t ?? '', filteredLinkResults);
+		const intervalId = setInterval(() => {
+			editSendMessage(result ? {} : filteredContent, idNewMessageResponse, filteredMention, filteredLinkResults);
+			setFilterContent({});
+			setFilteredLinkResults([]);
+			setFilterMention([]);
+			clearInterval(intervalId);
+		}, 1000);
```
nguyentrannhan, line 103: 1000 doi thanh 0 thoi chu

phong9xNcc: vâng ah

PR #2703, opened 2024-08-15
Add member into dm group

sadohpv:


libs/core/src/lib/chat/hooks/useFriends.ts
```diff
@@ -71,6 +72,13 @@ export function useFriends() {
 		[friends],
 	);
 
+	const addToDmGroupFriends = useCallback(() => {
```
nguyentrannhan, line 75: dday la ham list co phai add dau

libs/components/src/lib/components/DmList/CreateMessageGroup/index.tsx
```diff
@@ -50,14 +51,29 @@ const CreateMessageGroup = ({ onClose }: CreateMessageGroupProps) => {
 		setSelectedFriends([]);
 		onClose();
 	};
```
nguyentrannhan, line 53: xuong dong

libs/components/src/lib/components/DmList/DMtopbar/index.tsx
```diff
@@ -202,7 +195,32 @@ function PinButton({ isLightMode }: { isLightMode: boolean }) {
 		</div>
 	);
 }
```
nguyentrannhan, line 197: xuong dong

PR #2782, opened 2024-08-20
Feat Mobile: Search items settings

vyhuynhngoc:


apps/mobile/src/app/screens/settings/index.tsx
```diff
@@ -269,12 +272,40 @@ export const Settings = ({ navigation }: { navigation: any }) => {
...
+		setSearchText(text)
+
+		const results: IMezonMenuItemProps[] = [];
+		menu.forEach(section => {
+			if(!!section.title) {
+				const matchedItems = section.items.filter(item =>
+				  item.title.toLowerCase().includes(text.toLowerCase())
+				);
+				results.push(...matchedItems);
+			}
```
phuocnguyenncc: using debounce, so that functions `handleSearchChange` `renderedMenu` are not called multiple times when changing text

phuocnguyenncc (approved):

PR #2818, opened 2024-08-21
hotfix scroll when chat 

sadohpv:


libs/store/src/lib/channelmembers/channel.members.ts
```diff
@@ -338,7 +341,12 @@ export const channelMembers = createSlice({
 			.addCase(fetchChannelMembers.rejected, (state: ChannelMembersState, action) => {
 				state.loadingStatus = 'error';
 				state.error = action.error.message;
-			});
+			})
+      .addCase(updateCustomStatus.fulfilled, (state:ChannelMembersState,action)=>{
```
nguyentrannhan: state: ChannelMembersState, action

sadohpv: Đã fix ạ

PR #2834, opened 2024-08-21
remove memberVoiceChannel not use

sadohpv:


libs/components/src/lib/components/UserListVoiceChannel/UserListItemVoiceChannel.tsx
```diff
@@ -1,20 +1,15 @@
-import { useMembersVoiceChannel } from '@mezon/core';
+
```
phuocnguyenncc, line 1: `import { useMembersVoiceChannel } from '@mezon/core';`
search all xóa giúp anh bên folder mobile luôn với
filename: `ChannelListUserVoiceItem`

sadohpv: Đã xoá ạ

phuocnguyenncc (approved):

PR #2863, opened 2024-08-22
Fix keeping reply banner when switching the messages

anbuihoang:
Fix keeping reply banner when switching the messages

### DIFF
--- a/libs/components/src/lib/components/MessageBox/ReactionMentionInput/index.tsx
+++ b/libs/components/src/lib/components/MessageBox/ReactionMentionInput/index.tsx
@@ -159,23 +159,19 @@ function MentionReactInput(props: MentionReactInputProps): ReactElement {
 		[currentChannel?.channel_id, currentDmId, props.mode],
 	);
 
+	const userProfile = useSelector(selectAllAccount);
 	const idMessageRefEdit = useSelector(selectIdMessageRefEdit);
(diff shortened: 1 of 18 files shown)


libs/components/src/lib/components/MessageBox/ReactionMentionInput/index.tsx
```diff
@@ -157,30 +154,24 @@ function MentionReactInput(props: MentionReactInputProps): ReactElement {
...
-	const isSearchMessage = useSelector(
-		selectIsSearchMessage((props.mode === ChannelStreamMode.STREAM_MODE_CHANNEL ? currentChannel?.channel_id : currentDmId) || ''),
-	);
-	const { setDataReferences, setOpenThreadMessageState, setAttachmentData } = useReference(
-		(props.mode === ChannelStreamMode.STREAM_MODE_CHANNEL ? currentChannel?.channel_id : currentDmId) || '',
-	);
-	const attachmentDataRef = useSelector(
-		selectAttachmentData((props.mode === ChannelStreamMode.STREAM_MODE_CHANNEL ? currentChannel?.channel_id : currentDmId) || ''),
+	const currentDmOrChannelId = useMemo(
+		() => (props.mode === ChannelStreamMode.STREAM_MODE_CHANNEL ? currentChannel?.channel_id : currentDmId),
```
nguyentrannhan, line 158: switch case

Issue #2690, opened 2024-08-15
Mobile App: Thread creation from > The Thread name text field is hidden by the keyboard


trucnguyen31:
## Steps:

1. Hold on a specific channel > Select Thread option
2. At the Thread page, tap on plus icon 
3. At the Thread creation page, tap on the thread name text box
4. Observe

## ER:
![Image](https://github.com/user-attachments/assets/29026d7f-2e68-434f-8009-66c057033598)

## AR:
The Thread name text field is hidden by the keyboard
![Image](https://github.com/user-attachments/assets/d4e43313-1a56-4e56-82f1-2844df760210)

trucnguyen31:
@codengoo recheck

tien-cao:
PASSED

![Image](https://github.com/user-attachments/assets/97037fde-59f4-4f27-96f0-5beedb5c6f36)

PR #2867, opened 2024-08-22
Fix mobile: Keyboard hidden in create thread screen

codengoo:
[Mobile App: Thread creation from > The Thread name text field is hidden by the keyboard](https://github.com/nccasia/mezon-fe/issues/2690)
#2690


codengoo:
@phuocnguyenncc anh check hộ em trên ios có work không nhé ạ

phuocnguyenncc:
k work r. này để a check

PR #2921, opened 2024-08-26
Role UI short user and clan setting

sadohpv:


libs/components/src/lib/components/ClanSettings/SettingRoleManagement/SettingOptionRole/index.tsx
```diff
@@ -37,31 +41,32 @@ const SettingValueDisplayRole = ({ RolesClan}: { RolesClan: RolesClanEntity[]})
 			default:
 				return null;
 		}
-	  }, [selectedButton, RolesClan, hasPermissionEdit]);
+	  }, [selectedButton, RolesClan, hasPermissionEdit,clickRole]);
```
nguyentrannhan: space

sadohpv: Đã fix ạ

PR #2946, opened 2024-08-26
add showfilename

phong9xNcc:


libs/transport/src/lib/minio/index.ts
```diff
@@ -110,7 +110,7 @@ export async function handleUploadFileMobile(
 
 export function createUploadFilePath(session: Session, currentClanId: string, currentChannelId: string, filename: string): string {
 	const ms = new Date().getMinutes();
-	filename = ms + filename;
```
nguyentrannhan, line 113: cho nay truyen filename vao attachment

libs/components/src/lib/components/MessageWithUser/MessageLinkFile.tsx
```diff
@@ -53,7 +53,7 @@ function MessageLinkFile({ attachmentData }: MessageImage) {
 			<div className="flex items-center">{thumbnailAttachment}</div>
 			{hideTheInformationFile && (
 				<div className=" cursor-pointer " onClick={handleDownload} onKeyDown={handleDownload}>
-					<p className="text-blue-500 hover:underline">{attachmentData.filename}</p>
+					<p className="text-blue-500 hover:underline">{attachmentData.filename?.slice(attachmentData.filename?.lastIndexOf('/') + 1)}</p>
```
nguyentrannhan: truyen lam 2 params rieng biet la filename voi url

libs/transport/src/lib/minio/index.ts
phong9xNcc: fixed:
1. exported filePath && originalFilename
image
2. set filename == original name
image

libs/components/src/lib/components/MessageWithUser/MessageLinkFile.tsx
phong9xNcc: image
removed detect filename form  fullfillname

PR #2966, opened 2024-08-27
fix: display default notification + add enum

hmquannnnn:


libs/components/src/lib/components/ChannelTopbar/TopBarComponents/NotificationSetting/index.tsx
```diff
@@ -15,6 +15,7 @@ import { useEffect, useMemo, useState } from 'react';
 import { useSelector } from 'react-redux';
 import { notificationTypesList, notiLabels } from "../../../PanelChannel";
 import ItemPanel from "../../../PanelChannel/ItemPanel";
+import {ENotificationTypes} from "@mezon/utils";
```
nqdd: check format

hmquannnnn: done ạ

PR #2987, opened 2024-08-27
Feat jump to new message

nqdd:


nqdd:
implement for [#659](https://github.com/nccasia/mezon-fe/issues/659)

PR #2994, opened 2024-08-28
fix panel position, add esc, enter in category setting

hmquannnnn:
**mouseY = event.clientY + window.screenY** will cause mouseY to be calculated incorrectly when the browser window is minimized and moved downward.


libs/components/src/lib/components/CategorySetting/OverviewSetting.tsx
```diff
@@ -1,75 +1,93 @@
...
+	useEscapeKey(() => {
+		if (hasChanged) {
+			handleReset();
+		} else {
+			onClose();
+		}
+	});
+
+	const handlePressEnter = (e: React.KeyboardEvent<HTMLInputElement>) => {
+		if (e.key === 'Enter' && !isInvalidName) {
```
nqdd: delcare enum
https://gist.github.com/cjcliffe/1185173

libs/components/src/lib/components/CategorySetting/OverviewSetting.tsx
```diff
@@ -1,75 +1,93 @@
...
+	}, [categoryName, category?.category_name]);
+	const dispatch = useAppDispatch();
+	const currentClan = useSelector(selectCurrentClan);
+
+	const handleChangeCategoryName = (categoryName: string) => {
+		setCategoryName(categoryName);
+
+		const regex = ValidateSpecialCharacters();
+
+		if (categoryName.length === 0 || categoryName.length === 64 || !regex.test(categoryName)) {
```
nqdd: khai bao constant cho 0 va 64
CATEGORY_MIN_LENGTH
CATEGORY_MAX_LENGTH

libs/components/src/lib/components/CategorySetting/OverviewSetting.tsx
hmquannnnn: done ạ

libs/components/src/lib/components/CategorySetting/OverviewSetting.tsx
hmquannnnn: done ạ

nqdd:
ok

PR #3001, opened 2024-08-28
Fix not supporting mov file

levanthanh1907:


libs/components/src/lib/components/MessageWithUser/MessageAttachment.tsx
```diff
@@ -19,7 +19,10 @@ const classifyAttachments = (attachments: ApiMessageAttachment[]) => {
 	const documents: ApiMessageAttachment[] = [];
 
 	attachments.forEach((attachment) => {
-		if (attachment.filetype?.indexOf('video/mp4') !== -1 && !attachment.url?.includes('tenor.com')) {
+		if (
```
nqdd, line 22: nên khai báo enum cho MIME type
https://gist.github.com/hiaux0/a26b964f69955d7d3c2907b1e48788f7

PR #3004, opened 2024-08-28
fix: desktop notification

nqdd:
https://github.com/MatthieuLemoine/electron-push-receiver outdated for FCM
https://firebase.google.com/support/faq#fcm-depr-features

solution: create new push receiver base on https://github.com/javajuice1337/push-receiver-v2


apps/desktop/src/environments/environment.prod.ts
```diff
@@ -3,5 +3,15 @@ declare const __BUILD_VERSION__: string;
 export const environment = {
 	production: true,
 	version: __BUILD_VERSION__,
-	senderId: '285548761692'
+	senderId: '285548761692',
```
nguyentrannhan, line 6: sao ko lay trong .env chu?

PR #3052, opened 2024-08-29
Fix error 400 when sending a link - fix redirect on the web when joining the clan via the invitation link

anbuihoang:
Fix error 400 when sending a link - fix redirect on the web when joining the clan via the invitation link


libs/transport/src/lib/minio/index.ts
```diff
@@ -179,36 +179,44 @@ export function handleUrlInput(url: string): Promise<ApiMessageAttachment> {
...
-								width: 0,
-								height: 0
-							});
-						}
-					} else {
-						resolve(defaultAttachment);
-					}
-				})
-				.catch((e) => {
+	const typeImages = ['.jpg', '.jpeg', '.png', '.webp'];
```
nguyentrannhan: thiếu gif, apng...
https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types

anbuihoang: Em thêm rồi ạ

PR #3262, opened 2024-09-06
pin sticker and image

sadohpv:


libs/components/src/lib/components/ChannelTopbar/TopBarComponents/PinnedMessages/ItemPinMessage.tsx
```diff
@@ -85,4 +89,45 @@ const ItemPinMessage = (props: ItemPinMessageProps) => {
 	);
 };
 
+const ListAttachMent = ({ attachments }: { attachments: ApiMessageAttachment[] }) => {
```
nguyentrannhan: sao Ment lai viet hoa??

sadohpv: Đã fix ạ

PR #3266, opened 2024-09-07
fix @here and remove mention everyone

Cut666:


libs/utils/src/lib/constant/index.ts
```diff
@@ -13,6 +13,7 @@ export const KEY_KEYBOARD = { BACKSPACE: 8, TAB: 9, ENTER: 13, ESC: 27, UP: 38,
 export const GROUP_CHAT_MAXIMUM_MEMBERS = 9;
 export const EVERYONE_ROLE_ID = '1825450696619593728';
 export const EVERYONE_ROLE_TITLE = 'Every One';
```
nguyentrannhan: Everyone chu?

PR #3306, opened 2024-09-09
fix: jump to pin message

nqdd:


nqdd:
fix issue https://github.com/nccasia/mezon-fe/issues/2941
fix cannot load more message (same root cause)

PR #3317, opened 2024-09-09
Mobile: update logic get members channel base-on web

vyhuynhngoc:
Mobile App: update logic get members channel base-on web
[#3304](https://github.com/nccasia/mezon-fe/issues/3304)


apps/mobile/src/app/components/MemberStatus/index.tsx
```diff
@@ -25,9 +25,9 @@ enum EActionButton {
 export const MemberListStatus = React.memo(() => {
 	const { themeValue } = useTheme();
 	const styles = style(themeValue);
-	const currentChannel = useContext(threadDetailContext);
+	const currentChannel = useAppSelector(selectCurrentChannel);
```
phuocnguyenncc: re-check on DM

PR #3715, opened 2024-09-23
optimize: remove func un-used and fix issue upload and fix jump from noti fcm, scroll active channel

phuocnguyenncc:


phuocnguyenncc:
re-reck android

PR #3886, opened 2024-09-27
Mobile: optimize list members when press channel detail

hoangtranlehuy:
Optimize list members when press channel detail, change map to flashlist
https://github.com/nccasia/mezon-fe/issues/3878

### DIFF
--- a/apps/mobile/src/app/components/MemberStatus/index.tsx
+++ b/apps/mobile/src/app/components/MemberStatus/index.tsx
@@ -1,10 +1,11 @@
 import { BottomSheetModal } from '@gorhom/bottom-sheet';
 import { useChannelMembersOnlineStatus } from '@mezon/core';
 import { Icons } from '@mezon/mobile-components';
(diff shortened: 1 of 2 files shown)


apps/mobile/src/app/components/MemberStatus/index.tsx
```diff
@@ -93,36 +111,25 @@ export const MemberListStatus = React.memo(() => {
...
-									currentChannel={currentChannel}
-									isDMThread={isDMThread}
-								/>
-							))}
+							<FlashList
+								data={onlineMembers}
+								keyExtractor={(user, index) => `channelOnlineMember[${user?.id}][${index}]`}
+								renderItem={renderMemberItem}
+								estimatedItemSize={size.s_80}
+							/>
```
phuocnguyenncc: add props nestedScrollEnabled

phuocnguyenncc (approved):

PR #3978, opened 2024-10-02
convert GetPermissionByRoleIdChannelId, ListUserPermissionInChannel s…

thanhlong1203:
…ocket to api


libs/components/src/lib/components/ChannelSetting/Component/PermissionsChannel/PermissionManage/ListRoleMember.tsx
```diff
@@ -42,15 +42,17 @@ const ListRoleMember = memo((props: ListRoleMemberProps) => {
 					permissionRoleChannelActions.fetchPermissionRoleChannel({
 						channelId: channelId,
 						roleId: listManageInChannel[0].id,
-						userId: ''
```
nguyentrannhan: Sao cho nay phai cho nocache the?

PR #4219, opened 2024-10-14
Join to thread

phong9xNcc:


libs/components/src/lib/components/ChannelTopbar/index.tsx
```diff
@@ -143,6 +147,23 @@ function ThreadButton({ isLightMode }: { isLightMode: boolean }) {
 		setIsShowThread(false);
 	}, []);
 
+	const currentChannel = useSelector(selectCurrentChannel);
+	const isThread = currentChannel?.parrent_id !== '0';
```
nguyentrannhan, line 151: const isThread = currentChannel?.parrent_id !== '0' && currentChannel?.parrent_id !== '';

PR #4307, opened 2024-10-17
Fix list image show attachments

sadohpv:


libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -224,6 +225,17 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children }) =
 				const idToCompare = !isMobile ? channelId : currentChannelId;
 				mess.isCurrentChannel = message.channel_id === idToCompare;
 			}
+
+			if (
+				mess.attachments?.length &&
+				mess.attachments.length > 0 &&
+				mess.attachments.some((att) => att?.filetype?.startsWith(ETypeLinkMedia.IMAGE_PREFIX))
+			) {
+				dispatch(
```
huylexuan44: i think you should check add one in this case

apps/chat/src/app/loaders/channelLoader.ts
```diff
@@ -12,6 +12,8 @@ export const channelLoader: CustomLoaderFunction = async ({ params, request, dis
 
 	dispatch(channelsActions.joinChannel({ clanId: clanId ?? '', channelId: channelId, noFetchMembers: false, messageId: messageId || '' }));
 	dispatch(channelsActions.setPreviousChannels({ channelId: channelId }));
+	dispatch(attachmentActions.fetchChannelAttachments({ clanId: clanId ?? '', channelId }));
```
nguyentrannhan: Sao fetch cho nay chu?

libs/store/src/lib/attachment/attachments.slice.ts
```diff
@@ -104,7 +115,7 @@ export const attachmentSlice = createSlice({
 				state.loadingStatus = 'loading';
 			})
 			.addCase(fetchChannelAttachments.fulfilled, (state: AttachmentState, action: PayloadAction<any>) => {
-				attachmentAdapter.setAll(state, action.payload);
+				attachmentAdapter.setMany(state, action.payload);
```
nguyentrannhan: cho nay can than

huylexuan44 (approved):

PR #4781, opened 2024-11-08
update UI stream

loi-huynhphuc:
Screenshot 2024-11-08 165129
Screenshot 2024-11-08 163203


phuocnguyenncc:
@loi-huynhphuc crash 
image

PR #5197, opened 2024-12-03
update call log

levanthanh1907:


libs/utils/src/lib/types/index.ts
```diff
@@ -1295,3 +1295,16 @@ export type IPermissonMedia = 'granted' | 'denied' | 'not_found';
 export enum AttachmentTypeUpload {
 	BLOB = 'blob:'
 }
+
+export enum CallLog {
+	MISSED = 'Bạn bị nhỡ',
```
nguyentrannhan: để tiếng anh

phuocnguyenncc: a @levanthanh1907 "callLog": {
        "cancel": "You canceled",
        "missed": "You Missed",
        "receiverRejected": "Receiver Rejected",
        "youRejected": "You Rejected",
        "audioCall": "Audio call",
        "videoCall": "Video call",
        "callBack": "Call back",
        "incomingCall": "Incoming call",
        "outGoingCall": "Outgoing call"
    }

phuocnguyenncc (approved):

PR #5382, opened 2024-12-13
improve open image viewer

hmquannnnn:
send clicked image khi openImageWindow, sau đó mới fetch attachment list và send cho image viewer window


apps/desktop/src/app/api/main.preload.ts
huylexuan44, line 1: pull code develop and update

huylexuan44 (changes requested):

huylexuan44 (approved):

PR #5425, opened 2024-12-18
implement init message to the top of topic, add image proxy for view topic avatar, fix permission denied of "load more message" function in topic board

Tungnqs:


libs/components/src/lib/components/MessageWithUser/MessageContent.tsx
```diff
@@ -69,6 +81,13 @@ const MessageContent = ({ message, mode, isSearchMessage }: IMessageContentProps
 		[message]
 	);
 
+	const avatarToDisplay = useMemo(() => {
+		if (topicCreator?.clan_avatar) {
+			return topicCreator?.clan_avatar;
+		}
```
huylexuan44: remove use memo -> cost of use memo larger than assign directly

Tungnqs: fixed

huylexuan44 (approved):

PR #5472, opened 2024-12-22
realtime list topic

Cut666:


apps/chat/src/app/layouts/AppLayout.tsx
```diff
@@ -120,7 +120,14 @@ const AppLayout = () => {
 		)
 			.then((response): void => {
 				const token = (response?.payload as { token: string })?.token;
-				notificationService.connect(token);
+				notificationService.connect(token, (msg) => {
+					if (msg?.extras?.topicId && msg?.extras?.topicId !== '0') {
```
huylexuan44: check current view is channel of this topic,  and check exists to not handle

huylexuan44 (approved):

PR #5500, opened 2024-12-25
fix onboarding create and edit at the same time

sadohpv:


libs/store/src/lib/onboarding/onboarding.slice.ts
```diff
@@ -88,10 +88,10 @@ export const createOnboardingTask = createAsyncThunk(
 				clan_id,
 				contents: [...content]
 			});
-			if (!response) {
+			if (!response || !response.list_onboarding) {
```
huylexuan44: update to response || !response.list_onboarding =>  !response.?list_onboarding

huylexuan44 (approved):

PR #5557, opened 2024-12-30
update modal e2ee

levanthanh1907:


libs/components/src/lib/components/ModalSendCode/index.tsx
```diff
@@ -151,6 +151,7 @@ const ModalConfirmPin = ({ onClose, onBack, pin, userProfile }: ModalProps & { p
 	const [errorMessage, setErrorMessage] = useState<string>('');
 	const inputsRef = useRef<(HTMLInputElement | null)[]>([]);
 	const dispatch = useAppDispatch();
+	const { directId } = useAppParams();
```
huylexuan44, line 154: use selector currentDmId

PR #5577, opened 2024-12-31
remove logo

Cut666:


apps/chat/src/app/pages/main/index.tsx
```diff
@@ -444,7 +464,7 @@ const SidebarMenu = memo(
 								draggable="false"
 							>
 								<NavLinkComponent active={!isClanView}>
-									<div>
+									<div onContextMenu={handleMouseClick}>
```
huylexuan44: break to new component wrap logic,  keep main not rerender when handle event this component

huylexuan44 (approved):

PR #5607, opened 2025-01-03
fix member page when join stream

sadohpv:


apps/chat/src/app/layouts/ClanLayout.tsx
huylexuan44, line 1: update for all case use same logic

huylexuan44 (approved):

PR #5608, opened 2025-01-03
fix forward all not display + only forward message content

hmquannnnn:


libs/components/src/lib/components/ForwardMessage/index.tsx
```diff
@@ -64,13 +66,14 @@ const ForwardMessageModal = ({ openModal }: ModalParam) => {
 	const modeResponsive = useSelector(selectModeResponsive);
 	const membersInClan = useAppSelector((state) => selectAllChannelMembers(state, currentChannelId as string));
 	const isForwardAll = useSelector(getIsFowardAll);
+	const allMessageIds = useAppSelector((state) =>
```
huylexuan44: move select state (allMessageIds, allMessagesEntities) to inside function, it only use handle logic, not use to display in UI
and startIndex handleForwardAllMessage

huylexuan44: use const store = useStore();

huylexuan44 (approved):

PR #5646, opened 2025-01-07
fix the error when in this channel but setting in another channel, th…

thanhlong1203:
…e list user in channel not reset


huylexuan44:
need check impact

huylexuan44 (approved):

PR #5705, opened 2025-01-10
fixed bug mention/hashtash/emoji on markdown

phong9xNcc:


libs/core/src/lib/chat/hooks/useCheckTokenOnMarkdown.ts
```diff
@@ -0,0 +1,37 @@
+import { IEmojiOnMessage, IHashtagOnMessage, IMarkdownOnMessage, IMentionOnMessage } from '@mezon/utils';
+
+export const useCheckTokenOnMarkdown = (
```
huylexuan44: move to utils functions, it is not hook, and remove use prefix

huylexuan44 (approved):

PR #5708, opened 2025-01-10
fix Emoji Panel is covered when opening topic

levanthanh1907:


libs/components/src/lib/components/ChannelTopbar/index.tsx
```diff
@@ -103,10 +105,20 @@ const TopBarChannelVoice = memo(({ channel }: ChannelTopbarProps) => {
 
 const TopBarChannelText = memo(({ channel, isChannelVoice, mode, isMemberPath }: ChannelTopbarProps) => {
 	const dispatch = useAppDispatch();
+	const isShowCreateThread = useSelector((state) => selectIsShowCreateThread(state, channel?.id as string));
```
huylexuan44: use usestore

huylexuan44 (approved):

PR #5764, opened 2025-01-15
update activity

levanthanh1907:


apps/desktop/src/app/app.ts
```diff
@@ -264,53 +258,46 @@ export default class App {
...
-			console.error(ex);
-		}
+		const getActivityType = (appName: string): number | null => {
+			if (ACTIVITY_CODING.includes(appName as EActivityCoding)) return 1;
+			if (ACTIVITY_MUSIC.includes(appName as EActivityMusic)) return 2;
+			if (ACTIVITY_GAMING.includes(appName as EActivityGaming)) return 3;
+			return null;
+		};
+
+		fetchActiveWindow();
```
nguyentrannhan: Add try catch

apps/desktop/src/app/app.ts
```diff
@@ -264,53 +258,46 @@ export default class App {
...
 		}
 
-		activityTimeout = setInterval(() => {
-			try {
-				fetchActiveWindow();
-			} catch (ex) {
-				console.error(ex);
-			}
-		}, usageThreshold);
+		activityTimeout = setInterval(fetchActiveWindow, usageThreshold);
```
nguyentrannhan: Add try catch

apps/desktop/src/app/app.ts
levanthanh1907: i have fixed

apps/desktop/src/app/app.ts
levanthanh1907: i have fixed

huylexuan44 (approved):

PR #5765, opened 2025-01-15
Fix bug Leave Clan

Cut666:


libs/components/src/lib/components/PanelClan/index.tsx
```diff
@@ -89,6 +100,22 @@ const PanelClan: React.FC<IPanelCLanProps> = ({ coords, clan, setShowClanListMen
 		handClosePannel();
 	};
 
+	const { userId } = useAuth();
+	const currentClanId = useSelector(selectCurrentClanId);
+	const [isShowLeaveClanPopup, setIsShowLeaveClanPopup] = useState(false);
```
huylexuan44, line 102: make thunk action handle this case

huylexuan44 (approved):

PR #5795, opened 2025-01-16
scale down image alone

sadohpv:


libs/components/src/lib/components/MessageWithUser/MessageAttachment.tsx
```diff
@@ -155,24 +155,32 @@ const designLayout = (
...
-		if (!images[0]?.height) {
+		if (!images[0]?.height || !images[0]?.width) {
 			listImageSize[0] = {
 				height: 150,
 				width: images[0].width || 0
 			};
 			return listImageSize;
-		}
-		if ((images[0]?.width || 0) > MAX_WIDTH_ALBUM_IMAGE) {
+		} else {
```
huylexuan44, line 164: const calculateImageDimensions = ({ height = 0, width = 0 }) => {
  const MAX_HEIGHT = 275;
  const MAX_WIDTH = 550;

  if (!height || !width) {
    return { height: 150, width: width || 0 };
  }

  const aspectRatio = width / height;
  
  let newHeight = Math.min(height, MAX_HEIGHT);
  let newWidth = newHeight * aspectRatio;

  if (newWidth > MAX_WIDTH) {
    newWidth = MAX_WIDTH;
    newHeight = MAX_WIDTH / aspectRatio;
  }

  return {
    height: Math.round(newHeight),
    width: Math.round(newWidth)
  };
};
clean code later

huylexuan44 (approved):

PR #5856, opened 2025-01-21
fix bug UI topic

levanthanh1907:


apps/chat/src/app/pages/channel/ChannelMessages.tsx
```diff
@@ -92,6 +92,7 @@ function ChannelMessages({
 	const store = useStore();
 	const appearanceTheme = useSelector(selectTheme);
 	const currentChannelId = useSelector(selectCurrentChannelId);
+	const allUserIds = useAppSelector((state) => selectAllChannelMemberIds(state, currentChannelId));
```
huylexuan44: check performance this selector, duplicate

levanthanh1907: i have fixed

huylexuan44 (approved):

PR #5861, opened 2025-01-22
remove selector in message avatar and message user's name component; add left border for thread box and topic box

Tungnqs:


libs/components/src/lib/components/MessageWithUser/MessageAvatar.tsx
```diff
@@ -30,6 +28,14 @@ const MessageAvatar = ({ message, mode, onClick }: IMessageAvatarProps) => {
 		message?.clan_avatar
 	);
 
+	const avatarUrl = useMemo(() => {
```
huylexuan44: remove use memo

libs/components/src/lib/components/MessageWithUser/MessageHead.tsx
```diff
@@ -56,6 +55,11 @@ const MessageHead = ({ message, mode, onClick }: IMessageHeadProps) => {
 		message?.sender_id ?? ''
 	);
 
+	const priorityName = useMemo(() => {
```
huylexuan44: remove use memo

huylexuan44 (approved):

PR #5925, opened 2025-01-24
fix the bug undefined topic creator avatar when edit original init message; add real-time interaction for editing original initial message of topic

Tungnqs:


libs/components/src/lib/components/MessageWithUser/MessageInput.tsx
```diff
@@ -104,9 +104,19 @@ const MessageInput: React.FC<MessageInputProps> = ({ messageId, channelId, mode,
 			lk: channelDraftMessage?.draftContent?.lk,
 			mk: channelDraftMessage?.draftContent?.mk,
 			vk: channelDraftMessage?.draftContent?.vk,
-			tp: channelDraftMessage?.draftTopicId
+			tp: channelDraftMessage?.draftTopicId,
+			cid: channelDraftMessage?.draftContent?.cid
 		};
-	}, [channelDraftMessage?.draftContent, messageId]);
+	}, [
+		channelDraftMessage?.draftContent?.cid,
```
huylexuan44: update only use channelDraftMessage?.draftContent

huylexuan44 (approved):

PR #5969, opened 2025-02-04
fix: thread pagination not displayed

hmquannnnn:
image


libs/components/src/lib/components/ChannelTopbar/index.tsx
```diff
@@ -128,6 +128,20 @@ const TopBarChannelText = memo(({ channel, isChannelVoice, mode, isMemberPath }:
...
+
+	const fetchThreads = async () => {
+		if (!channel?.channel_id) {
+			return;
+		}
+		const body = {
+			channelId: channel?.parrent_id !== '0' ? (channel?.parrent_id ?? '') : (channel?.channel_id ?? ''),
+			clanId: channel?.clan_id ?? '',
+			page: 1,
+			noCache: true
```
huylexuan44: recheck

PR #5991, opened 2025-02-04
handle sending transaction message after using "give coffee" feature

Tungnqs:
- After clicking at the "give coffee" button, send a direct message to an user with the code 11 - stand for Token sending (declared in Enum).
-  Then, prevent showing the message with code 11 in normal message line and catch it in another case to show with TokenTransactionMessage component
- Furthermore, disable the "edit/delete message" function of transaction message


libs/components/src/lib/components/Message/ChannelMessageOpt.tsx
```diff
@@ -197,9 +196,22 @@ const RecentEmoji: React.FC<RecentEmojiProps> = ({ message }) => {
...
 	const channel = useAppSelector((state) => selectChannelById(state, message.channel_id ?? '')) || {};
+	const { createDirectMessageWithUser } = useDirect();
+	const { sendInviteMessage } = useSendInviteMessage();
+
+	const sendNotificationMessage = useCallback(
+		async (userId: string) => {
+			const response = await createDirectMessageWithUser(userId);
+			if (response.channel_id) {
+				const channelMode = ChannelStreamMode.STREAM_MODE_DM;
+				sendInviteMessage('Tokens sent: 10,000₫', response.channel_id, channelMode, TypeMessage.SendToken);
```
huylexuan44: make const for 10,000

Tungnqs:
- add format money function for message string, integrate send transaction message in "send token" function

huylexuan44 (approved):

PR #6036, opened 2025-02-06
fix the bug: EVENT _ Users can access review without completing event information

Tungnqs:
- prevent moving to the third tab when current tab is the first tab and topic title is empty


libs/components/src/lib/components/ChannelList/EventChannelModal/ModalCreate/index.tsx
```diff
@@ -109,6 +109,10 @@ const ModalCreate = (props: ModalCreateProps) => {
 			return;
 		}
 
+		if (currentModal === 0 && number === 2 && !contentSubmit.topic) {
```
huylexuan44: add const , can not hard code

huylexuan44 (approved):

PR #6044, opened 2025-02-06
fix bug add and leave user GR

Cut666:


libs/components/src/lib/components/LeaveGroupModal/index.tsx
```diff
@@ -12,14 +20,21 @@ function LeaveGroupModal({ groupWillBeLeave, onClose, navigateToFriends }: Leave
...
 	const handleCheckboxChange = () => {
 		setIsChecked(!isChecked);
 	};
 
 	const handleLeaveAndClose = async (e: React.MouseEvent<HTMLDivElement>) => {
 		e.stopPropagation();
-		await dispatch(directActions.closeDirectMessage({ channel_id: groupWillBeLeave.channel_id }));
+		await dispatch(
+			channelMembersActions.removeMemberChannel({
+				channelId: groupWillBeLeave.channel_id || '',
```
huylexuan44: validate no value show error instead of use ''

huylexuan44 (approved):

PR #6059, opened 2025-02-07
fix: ctrl x when input has mentioned

hmquannnnn:


libs/components/src/lib/components/MessageBox/ReactionMentionInput/ReactionMentionInput.tsx
```diff
@@ -555,11 +557,11 @@ export const MentionReactInput = memo((props: MentionReactInputProps): ReactElem
 				...request,
 				valueTextInput: newValue,
 				content: newPlainTextValue,
-				mentionRaw: mentions
+				mentionRaw: newMentions
 			},
 			isNotChannel
 		);
-		if (mentions.some((mention) => mention.display === TITLE_MENTION_HERE)) {
+		if (newMentions.some((mention) => mention.display === TITLE_MENTION_HERE)) {
```
phuocnguyenncc: check valid data newMentions before call some method avoid app crash

phuocnguyenncc (approved):

PR #6087, opened 2025-02-10
Fix the bug: cannot edit category of channel

Tungnqs:
- Use selector to get real-time channel entity value to change dropdown values 
- Create changeCategoryOfChannel thunk function to call the changeChannelCategory api 
- Update channel entity state value and channel position/value in flat list


libs/store/src/lib/channels/channels.slice.ts
```diff
@@ -366,6 +367,39 @@ export const updateChannel = createAsyncThunk('channels/updateChannel', async (b
...
+export const changeCategoryOfChannel = createAsyncThunk('channels/changeCategoryOfChannel', async (request: IUpdateChannelRequest, thunkAPI) => {
+	try {
+		const mezon = await ensureSession(getMezonCtx(thunkAPI));
+		const state = thunkAPI.getState() as RootState;
+		const clanId = state.clans.currentClanId;
+		const response = await mezon.client.changeChannelCategory(mezon.session, request.category_id as string, {
+			channel_id: request.channel_id,
+			clan_id: clanId as string
+		});
+		if (response) {
```
huylexuan44: return early

libs/store/src/lib/channels/listChannelRender.slice.ts
```diff
@@ -110,6 +110,33 @@ export const listChannelRenderSlice = createSlice({
 				};
 			}
 		},
+		updateChannelPositionInRenderedList: (state, action: PayloadAction<{ channelId: string; clanId: string; categoryId: string }>) => {
+			const { channelId, clanId, categoryId } = action.payload;
+			if (state.listChannelRender[clanId]) {
```
huylexuan44: return early, keep small block

huylexuan44 (approved):

PR #6134, opened 2025-02-12
fix wrong count react message

trinhtruongthiphuong:
https://github.com/orgs/nccasia/projects/16/views/1?filterQuery=trinh&pane=issue&itemId=96889160


phuocnguyenncc (approved):

phuocnguyenncc (changes requested):
Check case reaction for the first time

PR #6143, opened 2025-02-12
fix create thread on new channel

sadohpv:


libs/components/src/lib/components/CreateChannelModal/CreateChannelModal.tsx
```diff
@@ -86,7 +86,8 @@ export const CreateNewChannelModal = () => {
 			channel_label: channelName,
 			channel_private: isPrivate,
 			category_id: currentCategory?.category_id,
-			...(isAppChannel && { app_url: appUrl })
+			...(isAppChannel && { app_url: appUrl }),
+      parrent_id : '0'
```
huylexuan44, line 90: add const for global type parrent id

huylexuan44 (approved):

PR #6156, opened 2025-02-13
add the condition of no clan joint to show first join popup; add new button in Microsoft download dropdown; fix wrong display bug of message in delete channel popup

Tungnqs:


apps/chat/src/app/pages/main/index.tsx
```diff
@@ -83,7 +83,8 @@ function MyApp() {
 	const { userProfile } = useAuth();
 	const calculateJoinedTime = new Date().getTime() - new Date(userProfile?.user?.create_time ?? '').getTime();
 	const isNewGuy = calculateJoinedTime <= TIME_OF_SHOWING_FIRST_POPUP;
-	const [isShowFirstJoinPopup, setIsShowFirstJoinPopup] = useState(isNewGuy);
+	const clans = useSelector(selectAllClans);
```
huylexuan44: make totalClanNumber selector instead of use selectAllClans.length ,  it trigger render many time

apps/chat/src/app/pages/homepage/mezonpage/index.tsx
```diff
@@ -210,7 +211,16 @@ export const DropdownButton: React.FC<DropdownButtonProps> = ({ platform, icon,
 				</div>
 			</button>
 			{isOpen && (
-				<div className="absolute mt-[8px] z-50 flex flex-col gap-1">
+				<div className="absolute z-50 flex flex-col gap-1 mt-1 left-[2px]">
+					<a className="cursor-pointer block " href={downloadUrl} target="_blank" rel="noreferrer">
+						<div className="bg-black py-1 max-md:py-[2px] px-[10px] max-md:px-2 w-[180px] max-md:w-[125px] flex items-center gap-[10px] rounded-md border-[1.5px] border-white">
```
huylexuan44, line 220: replace Div to inline Tag, html semantic rule: inline tag not wrap inner block tags

huylexuan44 (approved):

PR #6158, opened 2025-02-13
fix bug disconnect mezon meet when changing clan and show voice info when join

thanhlong1203:


apps/chat/src/app/layouts/MainLayout.tsx
```diff
@@ -122,6 +129,12 @@ const MainLayout = memo(
 					event.preventDefault();
 				}}
 			>
+				<div
+					className={`h-full ${!isChannelMezonVoice ? 'hidden' : ''} absolute bottom-0 right-0 z-30`}
+					style={{ width: 'calc(100% - 72px - 272px)' }}
+				>
+					<ChannelVoice channel={currentChannel ?? undefined} roomName={currentChannel?.meeting_code || ''} />
```
huylexuan44: break wrapper channelVoice component and move to logic, reduce render Outlet many time, when user change channel

apps/chat/src/app/pages/channel/ChannelVoice.tsx
```diff
@@ -51,27 +51,27 @@ import { useSelector } from 'react-redux';
...
+	const voiceInfo = useSelector(selectVoiceInfo);
 	const [loading, setLoading] = useState<boolean>(false);
 	const dispatch = useAppDispatch();
-	const currentChannelId = useSelector(selectCurrentChannelId);
 	const serverUrl = process.env.NX_CHAT_APP_MEET_WS_URL;
 	const showMicrophone = useSelector(selectShowMicrophone);
 	const showCamera = useSelector(selectShowCamera);
 	const isVoiceFullScreen = useSelector(selectVoiceFullScreen);
 	const { userProfile } = useAuth();
+	const currentClan = useSelector(selectCurrentClan);
```
huylexuan44, line 65: don't use selector when it not use render UI, get state from thunk action or store instance
search example: const handleJoinChannel = async () => {

huylexuan44: nhân tiện tìm trong component này các selector k dùng để check render UI luôn chú

libs/components/src/lib/components/VoiceInfo/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+					onClick={leaveVoice}
+				>
+					<Icons.EndCall className="w-5 h-5" />
+				</button>
+			</div>
+		</div>
+	);
+};
+
+export default VoiceInfo;
```
huylexuan44, line 73: add react.memo

huylexuan44 (approved):

PR #6210, opened 2025-02-18
Bugfix/channel is not compative with mode

nam-trinhhoai:
check update channel private/public event
remove if toggle to private except creator
add if toggle to public 
update if creator


libs/components/src/lib/components/ChannelList/ChannelListItem/index.tsx
```diff
@@ -71,6 +71,9 @@ const ChannelLinkContent: React.FC<ChannelLinkContentProps> = ({ channel, isActi
 	const handleOpenInvite = () => {
 		dispatch(clansActions.toggleInvitePeople({ status: true, channelId: channel.id }));
 	};
+	const checkPrivateChannel = useMemo(() => {
```
huylexuan44: remove usememo, cost of usememo greater than assign operator

huylexuan44 (approved):

PR #6219, opened 2025-02-19
fix: close topic when its init message is deleted

hmquannnnn:


libs/components/src/lib/components/DeleteMessageModal/ModalDeleteMess.tsx
```diff
@@ -43,15 +45,26 @@ const ModalDeleteMess = (props: ModalDeleteMessProps) => {
...
 	const handleAction = useCallback(() => {
 		if (isRemoveAttachmentNoContent) {
 			const remainingAttachments =
 				attachmentData && mess?.attachments && mess?.attachments.filter((attachment) => attachment.url !== attachmentData.url);
 			editSendMessage(mess.content, mess.id, mess.mentions ?? [], remainingAttachments, true, '');
-		} else if (removeLastFile) {
-			deleteSendMessage(mess.id);
 		} else {
-			deleteSendMessage(mess.id);
+			handleDeleteMessage();
```
huylexuan44: i think add await, and add loading status pending call API

huylexuan44: button modal disable and show loading icon when api pending

PR #6221, opened 2025-02-19
Fix welcome direct hard code

sadohpv:


libs/components/src/lib/components/MarkdownFormatText/ModalUnknowChannel.tsx
```diff
@@ -12,14 +12,28 @@ type ModalUnknowChannelProps = {
...
 	const dispatch = useDispatch();
 	const { onClose, isError = false } = props;
-	const { navigate, toChannelPage } = useAppNavigation();
+	const { navigate, toChannelPage, toMembersPage, toFriendPage } = useAppNavigation();
 	const resetErrorToastStatus = () => {
 		dispatch(toastActions.setErrorToastStatus(false));
 	};
 
+	const currentClanId = useSelector(selectCurrentClanId);
+	const welcomeChannelId = useSelector(state => selectWelcomeChannelByClanId(state, currentClanId as string))
```
huylexuan44: state not render should use getStoreAsync get directly in action

libs/core/src/lib/app/hooks/useAppNavigation.ts
```diff
@@ -16,6 +16,10 @@ export function useAppNavigation() {
 		return `/`;
 	}, []);
 
+  const toFriendPage = useCallback(() => {
+		return `/chat/direct/friends`;
```
huylexuan44: useAppNavigation make multiple instance, i think we can keep everything small, currently component trigger rerender it will re compare all action,
-> make toFriendPage is const, and use use navigate directly in component , not use useAppNavigation

libs/components/src/lib/components/MarkdownFormatText/ModalUnknowChannel.tsx
```diff
@@ -12,14 +13,32 @@ type ModalUnknowChannelProps = {
 function ModalUnknowChannel(props: ModalUnknowChannelProps) {
 	const dispatch = useDispatch();
 	const { onClose, isError = false } = props;
-	const { navigate, toChannelPage } = useAppNavigation();
+	const { toChannelPage, toMembersPage } = useAppNavigation();
```
huylexuan44: fix

libs/utils/src/lib/constant/index.ts
```diff
@@ -99,3 +99,4 @@ export const TOKEN_TO_AMOUNT = {
 export const ADD_ROLE_CHANNEL_STATUS = 'Add Role Channel';
 export const KOMU_CLAN_ID = '1779484504377790464';
 export const WELCOME_CHANNEL_ID = '1827883133219901440';
+export const FRIEND_PAGE_LINK = '/chat/direct/friends';
```
huylexuan44: make navigation.ts move to it

huylexuan44 (approved):

PR #6225, opened 2025-02-19
fix the bug: [DESKTOP] Content is not changed when switching to another page in Search Message

Tungnqs:
- add state to store keyword and value of searched request
- get the request state value to assign to initial value of search state for the first fetching

### DIFF
--- a/libs/components/src/lib/components/SearchMessageChannel/index.tsx
+++ b/libs/components/src/lib/components/SearchMessageChannel/index.tsx
@@ -65,7 +65,7 @@ const SearchMessageChannel = ({ mode }: SearchMessageChannelProps) => {
 		() => (mode === ChannelStreamMode.STREAM_MODE_CHANNEL ? (currentChannel?.id ?? '') : (currentDmGroupId ?? '')),
 		[mode, currentChannel, currentDmGroupId]
 	);
-	const searchedRequest = useSelector(selectSearchedRequestByChannelId(channelId));
+	const searchedRequest = useSelector((state) => selectSearchedRequestByChannelId(state, channelId));
 
 	const valueInputSearch = useSelector(selectValueInputSearchMessage(channelId));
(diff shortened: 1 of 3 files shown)


libs/store/src/lib/searchmessages/searchmessage.slice.ts
```diff
@@ -160,3 +166,6 @@ export const selectIsSearchMessage = createSelector(
 
 export const selectValueInputSearchMessage = (channelId: string) =>
 	createSelector(getSearchMessageState, (state) => state.valueInputSearch[channelId]);
+
+export const selectSearchedRequestByChannelId = (channelId: string) =>
```
huylexuan44: update selector wrap by createSelector, this is rule of redux

libs/store/src/lib/searchmessages/searchmessage.slice.ts
```diff
@@ -160,3 +166,6 @@ export const selectIsSearchMessage = createSelector(
 
 export const selectValueInputSearchMessage = (channelId: string) =>
```
huylexuan44: update selector wrap by createSelector, this is rule of redux

libs/components/src/lib/components/SearchMessageChannel/index.tsx
```diff
@@ -234,7 +238,7 @@ const SearchMessageChannel = ({ mode }: SearchMessageChannelProps) => {
 			const requestBody = { ...search, filters: requestFilter, from: currentPage, size: SIZE_PAGE_SEARCH };
 			fetchSearchMessages(requestBody);
 		}
-	}, [currentPage]);
+	}, [channelId, currentClanId, currentPage, fetchSearchMessages, search]);
```
huylexuan44: add debounce time to event user type

huylexuan44 (approved):

PR #6243, opened 2025-02-20
Mobile app: fix reaction style overflow mobile

hoangtranlehuy:
Mobile app: fix reaction style overflow mobile

### DIFF
--- a/apps/mobile/src/app/screens/home/homedrawer/components/MessageReaction/styles.ts
+++ b/apps/mobile/src/app/screens/home/homedrawer/components/MessageReaction/styles.ts
@@ -12,7 +12,7 @@ export const style = (colors: Attributes) =>
 		},
 		reactionSpace: {
 			marginBottom: size.s_6,
-			height: size.s_30
+			height: 'auto'
 		},
 		myReaction: {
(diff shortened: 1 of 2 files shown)


apps/mobile/src/app/screens/home/homedrawer/components/MessageReaction/styles.ts
```diff
@@ -12,7 +12,7 @@ export const style = (colors: Attributes) =>
 		},
 		reactionSpace: {
 			marginBottom: size.s_6,
-			height: size.s_30
+			height: 'auto'
```
phuocnguyenncc: revert

apps/mobile/src/app/screens/home/homedrawer/components/MessageReaction/styles.ts
```diff
@@ -28,7 +28,7 @@ export const style = (colors: Attributes) =>
 			gap: size.s_2,
 			padding: size.s_2,
 			borderRadius: 5,
-			height: '100%'
+			height: size.s_30
```
phuocnguyenncc, line 31: revert

PR #6271, opened 2025-02-24
Fix the bug: [DESKTOP] DM List Shows Incorrect User After Messaging Stranger; [DESKTOP] Highlight Mention Occasionally Lost

Tungnqs:
https://github.com/orgs/nccasia/projects/16/views/1?filterQuery=tung&pane=issue&itemId=97342423
https://github.com/orgs/nccasia/projects/16/views/1?filterQuery=tung&pane=issue&itemId=97379881

Method used:
- set all direct entities to channel entity state with the key of clan id "0" after fetching all direct messages
- create async thunk function to get a specific direct entity that has been set in channel state; In the socket event catching, check for the dm is created or not, then create a new one with sender id when the sender id is not the same with current user id. 
- If the sender is equal to current user id, get partner id in the dmEntity state that is get by message.channelId to create new dm


libs/store/src/lib/direct/direct.slice.ts
```diff
@@ -160,6 +160,7 @@ export const fetchDirectMessage = createAsyncThunk(
 				return -1;
 			});
 			const channels = sorted.map(mapDmGroupToEntity).filter((item) => item.active);
+			thunkAPI.dispatch(channelsActions.setChannelEntityListByClanId({ channels: channels, clanId: '0' }));
```
huylexuan44: Why do you have to set this state?

huylexuan44 (approved):

PR #6289, opened 2025-02-25
hide the setting items in user setting that haven't been integrated

Tungnqs:


libs/components/src/lib/components/SettingProfile/SettingItem/index.tsx
```diff
@@ -52,23 +52,23 @@ const SettingItem = ({ onItemClick, initSetting }: { onItemClick?: (settingName:
 					Profiles
 				</button>
 				<br />
```
huylexuan44: a nghĩ add div bọc ngoài add 1 class hidden thì có sao k

huylexuan44 (approved):

PR #6419, opened 2025-03-06
fix image sender avatar

sadohpv:


huylexuan44 (approved):

libs/utils/src/lib/utils/index.ts
```diff
@@ -1080,7 +1080,9 @@ export const getAttachmentDataForWindow = (
 		return {
 			...image,
 			uploaderData: {
-				avatar: (uploader?.clan_avatar || uploader?.user?.avatar_url) as string,
+				avatar: (uploader?.clan_avatar ||
+					uploader?.user?.avatar_url ||
```
huylexuan44, line 1084: test desktop app

PR #6429, opened 2025-03-06
add notification division by category and load more page for each category 

thanhlong1203:


libs/components/src/lib/components/NotificationList/AllNotificationItem.tsx
```diff
@@ -54,30 +40,43 @@ function AllNotificationItem({ notify }: NotifyMentionProps) {
...
 	};
 
 	const allTabProps = {
-		message: parseNotify.content,
-		subject: parseNotify.subject,
-		code: parseNotify.code,
-		senderId: parseNotify.sender_id
+		message: {
+			...notify.content,
+			content: notify.content.content ? safeJSONParse(notify.content.content) : null,
```
huylexuan44: what's the problem with using JSONParser(notify) safe for all don't call multiple times

huylexuan44 (approved):

PR #6470, opened 2025-03-07
fixed error update the state of a component that has not been mounted

phong9xNcc:


libs/ui/src/lib/Image/index.tsx
```diff
@@ -10,6 +11,15 @@ export type ImageProps = ImgProps & {
 };
 
 function Image({ loading = 'lazy', ...rest }: ImageProps) {
+	const [isMounted, setIsMounted] = useState(false);
+
+	useEffect(() => {
+		setIsMounted(true);
+		return () => setIsMounted(false);
+	}, []);
```
huylexuan44: remove it and change  to ;

huylexuan44 (approved):

PR #6518, opened 2025-03-12
update version package

Cut666:


apps/mobile/src/app/screens/home/homedrawer/ChannelList/index.tsx
```diff
@@ -86,57 +85,55 @@ const ChannelList = () => {
 
 	const keyExtractor = useCallback((item, index) => item.id + item.isFavor?.toString() + index, []);
```
huylexuan44, line 86: why are you update mobile code ?

huylexuan44 (approved):

PR #6537, opened 2025-03-13
fix duplicate notification when change clan and some avatars on tab for you to change to anonymous

thanhlong1203:


libs/components/src/lib/components/NotificationList/index.tsx
```diff
@@ -77,11 +77,11 @@ function NotificationList({ rootRef }: NotificationProps) {
 
 		let category = null;
 
-		if (currentTabNotify === InboxType.INDIVIDUAL) {
+		if (currentTabNotify === InboxType.INDIVIDUAL && allNotificationForYou.data.length === 0) {
```
huylexuan44: make variable and check null allNotificationForYou?.data?.length === 0

huylexuan44 (approved):

PR #6545, opened 2025-03-13
fix show wrong avatar on reply

phong9xNcc:


libs/components/src/lib/components/MessageWithUser/index.tsx
```diff
@@ -139,6 +139,17 @@ function MessageWithUser({
 	const isDM = mode === ChannelStreamMode.STREAM_MODE_GROUP || mode === ChannelStreamMode.STREAM_MODE_DM;
 
 	const [isAnonymousOnModal, setIsAnonymousOnModal] = useState<boolean>(false);
+	const [clickedOnReply, setClickedOnReply] = useState<boolean>(false);
+
+	const openUserProfileOnReply = (e: React.MouseEvent<HTMLImageElement, MouseEvent>, messageSenderId: string, isAnonymousReplied: boolean) => {
+		handleOpenShortUser(e, messageSenderId, isAnonymousReplied);
+		setClickedOnReply(true);
```
huylexuan44: reduce logic

huylexuan44 (approved):

PR #6562, opened 2025-03-14
fix empty markdown

sadohpv:


libs/components/src/lib/components/MessageBox/ReactionMentionInput/ReactionMentionInput.tsx
```diff
@@ -325,6 +325,11 @@ export const MentionReactInput = memo((props: MentionReactInputProps): ReactElem
 			};
 			const checkedRequest = request ? request : emptyRequest;
 			const { text, entities } = parseHtmlAsFormattedText(hasToken ? checkedRequest.content : checkedRequest.content.trim());
+
+			if (!text) {
+				return;
```
huylexuan44: check send attachment

libs/components/src/lib/components/MessageBox/ReactionMentionInput/ReactionMentionInput.tsx
```diff
@@ -325,6 +325,11 @@ export const MentionReactInput = memo((props: MentionReactInputProps): ReactElem
...
+			}
+
```
huylexuan44: check all case send msg

huylexuan44 (approved):

PR #6585, opened 2025-03-17
recent emoji

Cut666:


libs/components/src/lib/components/Message/ChannelMessageOpt.tsx
```diff
@@ -92,7 +92,8 @@ const ChannelMessageOpt = ({
 	const giveACoffeeMenu = useGiveACoffeeMenuBuilder(message);
 	const checkMessageOnTopic = useAppSelector((state) => selectIsMessageChannelIdMatched(state, message?.channel_id ?? ''));
 	const checkMessageHasTopic = useAppSelector((state) => selectIsMessageChannelIdMatched(state, message?.topic_id ?? ''));
-	const doNotAllowCreateTopic = (isTopic && checkMessageOnTopic) || (isTopic && checkMessageHasTopic) || !hasPermission;
+	const [canSendMessage] = usePermissionChecker([EOverriddenPermission.sendMessage], currentChannel?.channel_id || '');
```
huylexuan44: canSendMessage get from prop not valid ?

libs/core/src/lib/chat/hooks/useChatReaction.ts
```diff
@@ -17,6 +17,7 @@ import {
 } from '@mezon/store';
 import { EmojiStorage, transformPayloadWriteSocket } from '@mezon/utils';
 import AsyncStorage from '@react-native-async-storage/async-storage';
+import { selectAllEmojiRecent, selectLastEmojiRecent } from 'libs/store/src/lib/emojiSuggestion/emojiRecent.slice';
```
huylexuan44: import not correct

libs/core/src/lib/chat/hooks/useChatReaction.ts
```diff
@@ -41,6 +42,8 @@ export function useChatReaction({ isMobile = false, isClanViewMobile = undefined
 	const thread = useSelector(selectThreadCurrentChannel);
 	const isFocusThreadBox = useSelector(selectClickedOnThreadBoxStatus);
 	const isFocusTopicBox = useSelector(selectClickedOnTopicStatus);
+	const lastEmojiRecent = useSelector(selectLastEmojiRecent);
```
huylexuan44: selector not render use getStore() to get state , move to inside function

huylexuan44 (approved):

PR #6616, opened 2025-03-18
Add foucus popup

phong9xNcc:


libs/components/src/lib/components/ChannelLink/ChannelLink.tsx
```diff
@@ -143,8 +145,10 @@ const ChannelLinkComponent = ({
 	};
 
 	const currentMission = useSelector((state) => selectCurrentMission(state, clanId as string));
-
+	const store = getStore();
 	const handleClick = () => {
+		const isChannelApp = channel.type === ChannelType.CHANNEL_TYPE_APP;
```
huylexuan44, line 150: const store = getStore(); move to handleClick

huylexuan44 (approved):

PR #6652, opened 2025-03-20
integrate "Ctrl + B" button group shortcut for text bolding in chat box; add style for bold text in mention input field

Tungnqs:
- declare new component field of mention with the child index "2", which is stand for bold mention
- in the on key down event, add new case of "ctrl + b/cmd + b"
- loop through the raw mention array of the request state to calculate the start position of selected text in the text values with ids field of the state.
- handle bolding text in the case: select over a mention, select inside a mention, select over 2 mentions, ...


libs/components/src/lib/components/MessageBox/ReactionMentionInput/ReactionMentionInput.tsx
```diff
@@ -126,7 +128,24 @@ type ChannelsMentionProps = {
 	subText: string;
 };
 
+function addSymbolsAndIdsLengthOfMention(value: number, mentionType: number) {
+	let newValue = value;
+	if (mentionType === ETypeMEntion.HASHTAG || mentionType === ETypeMEntion.MENTION) {
+		newValue += 23;
```
huylexuan44: define const, should not hard code

libs/components/src/lib/components/MessageBox/ReactionMentionInput/ReactionMentionInput.tsx
```diff
@@ -268,6 +287,83 @@ export const MentionReactInput = memo((props: MentionReactInputProps): ReactElem
 			}
 		}
 
+		if ((ctrlKey || metaKey) && (key === 'b' || key === 'B')) {
```
huylexuan44, line 275: break logic to new function handlerBoldKey

libs/components/src/lib/components/MessageBox/ReactionMentionInput/ReactionMentionInput.tsx
```diff
@@ -268,6 +287,83 @@ export const MentionReactInput = memo((props: MentionReactInputProps): ReactElem
...
+
+			const mentionItem: MentionItem = {
+				display: boldedText,
+				id: '',
+				childIndex: ETypeMEntion.BOLD,
+				index: valueStart,
+				plainTextIndex: plainTextStart
+			};
+
+			if (request.mentionRaw.length > 0) {
```
huylexuan44: self test send msg has mention has emoji  has link ...

huylexuan44 (approved):

PR #6657, opened 2025-03-21
realtime update user profile for DM

Cut666:


libs/store/src/lib/direct/direct.slice.ts
```diff
@@ -555,6 +555,33 @@ export const directSlice = createSlice({
...
+
+			const dmGroup = state.entities?.[dmId as string];
+
+			if (!dmGroup || !user_id) return;
+
+			dmGroup.user_id = dmGroup.user_id ?? [];
+
+			const index = dmGroup.user_id?.indexOf(user_id);
+
+			if (index !== undefined && index !== -1) {
```
huylexuan44: refactor function many block wrap to code , return early

huylexuan44 (approved):

PR #6697, opened 2025-03-24
fix wrong type array

sadohpv:


libs/store/src/lib/channels/channels.slice.ts
```diff
@@ -1598,7 +1598,7 @@ export const selectCurrentCategory = createSelector(
 
 export const selectAppChannelsListShowOnPopUp = createSelector(
 	[getChannelsState, (state: RootState) => state.clans.currentClanId as string],
-	(state, clanId) => Object.values(state.byClans[clanId]?.appChannelsListShowOnPopUp || {})
+	(state, clanId) => Object.values(state.byClans[clanId]?.appChannelsListShowOnPopUp || [])
```
huylexuan44, line 1601: fix root cause, set object not array

huylexuan44 (approved):

PR #6764, opened 2025-03-27
update fix link

phong9xNcc:


libs/components/src/lib/components/MarkdownFormatText/MarkdownContent.tsx
```diff
@@ -33,7 +34,7 @@ export const MarkdownContent: React.FC<MarkdownContentOpt> = ({
 	const basePath = '/chat/clans/';
 	const contentHasChannelLink = content?.includes(basePath) && content?.includes('/channels/');
 	const channelIdOnLink = contentHasChannelLink ? content?.split('/')?.[content?.split('/')?.indexOf('channels') + 1] : undefined;
-	const isLinkChannel = !!channelIdOnLink;
+	const channel = useTagById(channelIdOnLink);
```
huylexuan44: if(channelIdOnLink) {
    const store = getStore();
    const channel = useTagById(channelIdOnLink) -> update  logic useTagById to selector;
    instead of check all case with useTagById, only get by store instace
}

huylexuan44 (approved):

PR #6788, opened 2025-03-28
support click out side and esc for modals in events

thanhlong1203:
-support click out side and esc for modals in events
-add number of interested members
-fix location display error


libs/components/src/lib/components/ChannelList/EventChannelModal/ModalCreate/modalShareEvent.tsx
```diff
@@ -38,10 +38,30 @@ const ModalShareEvent = (props: ModalShareEventProps) => {
...
 	useOnClickOutside(panelRef, closeModal);
 
+	const handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+		if (e.key === 'Escape') {
+			e.preventDefault();
+			e.stopPropagation();
+			closeModal();
+		}
+	};
+
```
huylexuan44, line 51: update to use useEscapeKeyClose for all component

huylexuan44 (approved):

PR #6859, opened 2025-04-02
fix break line in embed

sadohpv:


libs/components/src/lib/components/EmbedMessage/components/EmbedDescription.tsx
```diff
@@ -25,7 +25,11 @@ export function EmbedDescription({ description }: EmbedDescriptionProps) {
 			const contentInElement = text?.substring(s, e);
 
 			if (lastindex < s) {
-				formattedContent.push(<p key={`plain-${lastindex}`}>{text?.slice(lastindex, s) ?? ''}</p>);
+				formattedContent.push(
+					<p key={`plain-${lastindex}`} style={{ whiteSpace: 'pre-line' }}>
```
huylexuan44: update use tailwind class

huylexuan44 (approved):

PR #6871, opened 2025-04-02
Feat/join call sound

hmquannnnn:


libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -195,8 +197,11 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children }) =
 				const store = getStore();
 				const state = store.getState();
 				const voiceChannel = selectChannelById(state, voice.voice_channel_id);
+				const memberList = selectVoiceChannelMembersByChannelId(state, voice.voice_channel_id);
+				const currentUserId = selectCurrentUserId(state);
+				const hasJoinSoundEffect = memberList.some(member => member.user_id === currentUserId) || currentUserId === voice.user_id;
 
-				if (voiceChannel.type === ChannelType.CHANNEL_TYPE_MEZON_VOICE) {
+				if (voiceChannel.type === ChannelType.CHANNEL_TYPE_MEZON_VOICE && hasJoinSoundEffect) {
 					const joinSoundElement = document.createElement('audio');
```
huylexuan44, line 205: make one element on index.html, don't need create and remove when user join and out call

huylexuan44 (approved):

libs/core/src/lib/chat/contexts/ChatContext.tsx
huylexuan44: you can update later, small issue

PR #6878, opened 2025-04-03
update scroll bar color

Tungnqs:


apps/chat/src/styles.scss
```diff
@@ -123,6 +123,11 @@
 	width: 8px;
 }
 
+::-webkit-scrollbar-thumb {
```
huylexuan44: check only apply for class, don't  use global style

huylexuan44 (approved):

PR #6880, opened 2025-04-03
updated only owner to see private ev

phong9xNcc:


libs/core/src/lib/chat/hooks/useEventManagementQuantity.ts
```diff
@@ -1,21 +1,23 @@
...
-import { useSelector } from 'react-redux';
 
 export const useEventManagementQuantity = () => {
-	const currentClanId = useSelector(selectCurrentClanId);
-	const allEventManagement = useAppSelector((state) => selectEventsByClanId(state, currentClanId as string));
-
-	const allThreadChannelPrivate = useSelector(selectAllTextChannel);
+	const store = getStore();
+	const userId = selectAllAccount(store.getState())?.user?.id;
+	const currentClanId = selectCurrentClanId(store.getState());
```
huylexuan44: update to use selector

huylexuan44 (approved):

PR #6891, opened 2025-04-03
fix the inappropriate last seen setting when jumping to old message

Tungnqs:


libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -538,7 +538,6 @@ export const jumpToMessage = createAsyncThunk(
 							messageId: messageId,
 							direction: Direction_Mode.AROUND_TIMESTAMP,
 							isFetchingLatestMessages,
-							isClearMessage: true,
```
huylexuan44, line 536: i will check impact

libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -1460,6 +1459,10 @@ export const selectMessageByMessageId = createSelector(
 	}
 );
 
+export const selectLastMessageStateByChannelId = createSelector([getMessagesState, (state, channelId: string) => channelId], (state, channelId) => {
```
huylexuan44: update name to lastseen

libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -533,7 +533,6 @@ export const jumpToMessage = createAsyncThunk(
 							messageId: messageId,
 							direction: Direction_Mode.AROUND_TIMESTAMP,
 							isFetchingLatestMessages,
-							isClearMessage: true,
```
huylexuan44, line 536: revert

huylexuan44 (approved):

PR #6938, opened 2025-04-08
Imp popout voice

phong9xNcc:


libs/components/src/lib/components/VoiceChannel/PopOut/VoicePopout.tsx
```diff
@@ -0,0 +1,97 @@
+/* eslint-disable no-console */
+import { LiveKitRoom } from '@livekit/components-react';
```
huylexuan44: move this component to src/chat/app/components, this component not shared components

huylexuan44 (approved):

PR #6990, opened 2025-04-14
 fix DM channel create and update

sadohpv:


libs/store/src/lib/direct/direct.slice.ts
```diff
@@ -162,6 +161,7 @@ export const fetchDirectMessage = createAsyncThunk(
 			});
 			const channels = sorted.map(mapDmGroupToEntity);
 			thunkAPI.dispatch(directMetaActions.setDirectMetaEntities(channels));
+			console.log('channels: ', channels);
```
nguyentrannhan: console log

libs/store/src/lib/direct/direct.slice.ts
```diff
@@ -299,6 +299,7 @@ export const joinDirectMessage = createAsyncThunk<void, JoinDirectMessagePayload
 );
 
 const mapMessageToConversation = (message: ChannelMessage): DirectEntity => {
+	console.log('message: ', message);
```
nguyentrannhan: console log

huylexuan44 (approved):

PR #7134, opened 2025-04-24
 Fix extend chat voice

sadohpv:


libs/components/src/lib/components/PreJoinCalling/index.tsx
```diff
@@ -87,6 +91,44 @@ export default function PreJoinCalling() {
 
 	const getExternalToken = useSelector(selectExternalToken);
 	const getJoinCallExtStatus = useSelector(selectJoinCallExtStatus);
```
huylexuan44, line 84: move this page to app/chat/pages, reduce total bundle size of main, it only use for external meet

libs/components/src/lib/components/PreJoinCalling/index.tsx
```diff
@@ -377,3 +369,60 @@ export default function PreJoinCalling() {
 		</div>
 	);
 }
+const ChatStream = () => {
```
huylexuan44: break new file

libs/components/src/lib/components/PreJoinCalling/index.tsx
```diff
@@ -87,6 +91,44 @@ export default function PreJoinCalling() {
...
+	function createGuestSessionData(token: string) {
+		const payload = decodeJWT(token);
+		const now = Math.floor(Date.now() / 1000);
+		return {
+			created: false,
+			token,
+			created_at: now,
+			expires_at: payload.exp,
+			refresh_expires_at: undefined,
+			username: payload.usn || payload.usr || payload.sub || 'guest',
```
huylexuan44: make const

huylexuan44 (approved):

PR #7185, opened 2025-05-02
Fixbug and invite after login

kind6688:


apps/chat/src/app/pages/invite/index.tsx
```diff
@@ -1,17 +1,26 @@
...
 
 export default function InvitePage() {
 	const { inviteId: inviteIdParam } = useParams();
 	const selectInvite = useSelector(selectInviteById(inviteIdParam || ''));
 	const navigate = useNavigate();
 	const { inviteUser } = useInvite();
+	const isLogin = useSelector(selectIsLogin);
+	const dispatch = useDispatch();
+
+	useEffect(() => {
```
huylexuan44: recheck, i think it not work,  handle logic in ProtectedRoutes

huylexuan44 (approved):

PR #7218, opened 2025-05-06
change gotifyToken by accessToken 

kind6688:
…nctions


apps/chat/src/app/pages/channel/ChannelStream.tsx
huylexuan44, line 1: rename branch

PR #7245, opened 2025-05-08
Mobile app: fix hashtag not map when create new thread mobile

hoangtranlehuy:
Mobile app: fix hashtag not map when create new thread mobile.
issue: https://github.com/orgs/nccasia/projects/16/views/1?filterQuery=hoang&pane=issue&itemId=109400209


hoangtranlehuy:
image

phuocnguyenncc (approved):

PR #7267, opened 2025-05-09
Fix remove selector flowbite

sadohpv:


apps/admin/src/app/components/InputField/CustomSelectField.tsx
```diff
@@ -19,23 +18,22 @@ function CustomSelectField({
 	return (
 		<div className="ImageField mt-2">
 			{label && <label className="block text-sm">{label}</label>}
-			<Select
+			<select
```
huylexuan44, line 21: run admin app and test

huylexuan44 (approved):

PR #7278, opened 2025-05-11
[WIP]:  functionality to chat in video conference

kind6688:


libs/components/src/lib/components/VoiceChannel/MyVideoConference/MyVideoConference.tsx
```diff
@@ -142,124 +146,399 @@ export function MyVideoConference({
 		};
 	}, []);
 
+	const room = useRoomContext();
```
huylexuan44, line 149: remove logic and flow ChatStream

PR #7338, opened 2025-05-14
 create group for clan

kind6688:


apps/chat/src/app/pages/main/index.tsx
huylexuan44, line 1: i will check and update logic

PR #7388, opened 2025-05-19
Mobile App: fix not show preview file image

thanhlichncc:
fix: not show preview file image

### DIFF
--- a/apps/mobile/src/app/utils/helpers.ts
+++ b/apps/mobile/src/app/utils/helpers.ts
@@ -26,6 +26,8 @@ export const isImage = (url?: string) => {
 	return /\.(jpg|jpeg|png|webp|avif|gif|svg|heic|PNG)/.test(url);
 };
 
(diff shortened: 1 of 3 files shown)


apps/mobile/src/app/utils/helpers.ts
```diff
@@ -26,6 +26,8 @@ export const isImage = (url?: string) => {
 	return /\.(jpg|jpeg|png|webp|avif|gif|svg|heic|PNG)/.test(url);
 };
 
+export const checkFileTypeImage = (type: string) => type.startsWith('image/');
```
phuocnguyenncc: Add `?` to fix crash in the case of undefine

thanhlichncc: dạ đã thêm, anh check lại giúp em ạ

phuocnguyenncc (approved):

PR #7536, opened 2025-05-30
fix logout auth multiple

sadohpv:


libs/store/src/lib/auth/auth.slice.ts
```diff
@@ -228,7 +228,7 @@ export const authSlice = createSlice({
 			state.isLogin = true;
 		},
 		setLogout(state) {
-			if (state.session && state.activeAccount && Object.keys(state.session).length >= 2) {
+			if (state.session && state.activeAccount && Object.keys(state.session).length >= 2 && !(state.session as any).token) {
```
huylexuan44: clean your code define condition variable

huylexuan44 (approved):

PR #7572, opened 2025-06-02
Mobile app: fix setup permission item toggle switch mobile

hoangtranlehuy:
Mobile app: fix setup permission item toggle switch mobile.
Issue: https://github.com/orgs/nccasia/projects/16/views/1?filterQuery=hoang&pane=issue&itemId=113243388
Expect: 
- Click to switch will change UI and state switch permission state.
- Click to item will toggle switch and change permission state.
- Show save button when have any change in permission list.
- Hide save button in init and when permission change is equal init permission state.
- Save and change permission list of role when click save button.

### DIFF
--- a/apps/mobile/src/app/utils/helpers.ts
+++ b/apps/mobile/src/app/utils/helpers.ts
@@ -117,3 +117,8 @@ export function combineMessageReactions(reactions: any[], message_id: string): a
 
 	return dataCombinedArray;
 }
+
+export function isEqualStringArrayUnordered(a: string[], b: string[]): boolean {
+	if (a.length !== b.length) return false;
+	return JSON.stringify([...a].sort()) === JSON.stringify([...b].sort());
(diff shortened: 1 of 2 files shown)


hoangtranlehuy:
https://github.com/user-attachments/assets/32608b36-859b-46f5-984b-543597e3cd58

apps/mobile/src/app/utils/helpers.ts
```diff
@@ -117,3 +117,8 @@ export function combineMessageReactions(reactions: any[], message_id: string): a
 
 	return dataCombinedArray;
 }
+
+export function isEqualStringArrayUnordered(a: string[], b: string[]): boolean {
```
phuocnguyenncc, line 121: wrap to try catch

phuocnguyenncc (approved):

PR #7603, opened 2025-06-04
Fix profile and show noti copy id app 

kind6688:


libs/components/src/lib/components/FooterProfile/index.tsx
```diff
@@ -240,21 +241,43 @@ function FooterProfile({ name, status, avatar, userId, isDM }: FooterProfileProp
 	const GroupCallJoined = useSelector(selectGroupCallJoined);
 	const statusMenu = useSelector(selectStatusMenu);
 
-	const [openProfileModal, closeProfileModal] = useModal(() => {
-		return (
-			<div ref={rootRef}>
-				<ModalFooterProfile
-					userId={userId ?? ''}
-					avatar={avatar}
```
huylexuan44: restore code

huylexuan44 (approved):

PR #7612, opened 2025-06-05
Fix emoji trunsfer transaction

kind6688:
**about Transaction test after edit UI:** 
_case 1: skeleton woking for History and detail Transaction
case 2: Coppy woking ID transaction_

**about Emoji:** 
_case1: after send mess , off panel emoji  
case2: keep emoji in chat and voice_

**about Transfer** 
_case 1: Transfer  money woking
case 2: cant anable Button Send if choose user from selectdown_


libs/components/src/lib/components/MessageBox/ReactionMentionInput/ReactionMentionInput.tsx
```diff
@@ -328,8 +331,8 @@ export const MentionReactBase = memo((props: MentionReactBaseProps): ReactElemen
 					anonymousMessage,
 					mentionEveryone
 				);
-
 				setMentionEveryone(false);
+				setSubPanelActive(SubPanelName.NONE);
```
huylexuan44: call once this function, move top top handleSend function

huylexuan44 (approved):

PR #7628, opened 2025-06-06
fix send token to bot not in clan

sadohpv:


libs/components/src/lib/components/ModalUserProfile/StatusProfile/ModalSendToken.tsx
```diff
@@ -150,7 +150,6 @@ const ModalSendToken = ({
 
 	const handleSendToken = () => {
 		const userData = mergedUsers.find((user) => user.id === selectedUserId);
-		if (!userData) return;
```
huylexuan44, line 153: recheck if userdata is undefined

PR #7629, opened 2025-06-06
Create Upload Sound UI

kind6688:


libs/ui/src/lib/Icons/icons.tsx
```diff
@@ -2903,6 +2903,28 @@ export function ImageUploadIcon(props: React.HTMLAttributes<SVGElement>) {
 		</svg>
 	);
 }
+export function UploadSoundIcon(props: React.HTMLAttributes<SVGElement>) {
+	return (
+		<svg
```
huylexuan44: remove, check it exists

libs/components/src/lib/components/ClanSettings/SettingSoundEffect/ModalUploadSound.tsx
```diff
@@ -0,0 +1,303 @@
+"use client"
```
huylexuan44: remove

huylexuan44 (approved):

PR #7632, opened 2025-06-06
fix cached user status

sadohpv:


libs/store/src/lib/userstatus/userstatusAPI.slice.ts
```diff
@@ -23,11 +24,25 @@ export const userStatusAdapter = createEntityAdapter({
 	selectId: (status: UserStatusEntity) => status.id || ''
 });
 
+export const getUserStatusCached = memoizeAndTrack(
```
huylexuan44, line 27: recheck, it call some API ?

libs/store/src/lib/clans/clans.slice.ts
```diff
@@ -121,10 +122,25 @@ export const changeCurrentClan = createAsyncThunk<void, ChangeCurrentClanArgs>(
 	}
 );
 
+const fetchClansCached = memoizeAndTrack(
```
huylexuan44, line 125: please test crud clan, and switch account

libs/store/src/lib/activities/activitiesAPI.slice.ts
```diff
@@ -42,11 +43,25 @@ export const createActivity = createAsyncThunk('activity/createActiviy', async (
 	}
 });
 
+const listActivitiesCached = memoizeAndTrack(
```
huylexuan44, line 46: why activity call this time ?

huylexuan44 (approved):

PR #7659, opened 2025-06-09
fix bug hidden notisetting inthread

kind6688:


libs/components/src/lib/components/PanelChannel/index.tsx
```diff
@@ -386,7 +386,9 @@ const PanelChannel = ({ coords, channel, openSetting, setIsShowPanelChannel, onD
 							<ItemPanel children={nameChildren} onClick={() => muteOrUnMuteChannel(1)} subText={mutedUntil} />
 						)}
 
-						{channel?.type === typeChannel.text && (
+						{(channel && channel.type !== undefined &&
+							(channel.type === typeChannel.text || channel.type === typeChannel.thread ||
```
huylexuan44: make const for condition to read able

libs/components/src/lib/components/PanelChannel/index.tsx
```diff
@@ -468,7 +470,9 @@ const PanelChannel = ({ coords, channel, openSetting, setIsShowPanelChannel, onD
 						) : (
 							<ItemPanel children={nameChildren} onClick={() => muteOrUnMuteChannel(1)} subText={mutedUntil} />
 						)}
-						{(channel.type === typeChannel.text || channel.type === typeChannel.thread) && (
+							{(channel && channel.type !== undefined &&
+								(channel.type === typeChannel.text || channel.type === typeChannel.thread ||
```
huylexuan44: make const for condition to read able

huylexuan44 (approved):

PR #7662, opened 2025-06-10
fix dubplicate user in voice

sadohpv:


libs/store/src/lib/voice/voice.slice.ts
```diff
@@ -290,11 +292,12 @@ export const voiceActions = {
  *
  * See: https://react-redux.js.org/next/api/hooks#useselector
  */
-const { selectAll } = voiceAdapter.getSelectors();
+const { selectAll, selectEntities } = voiceAdapter.getSelectors();
 
 export const getVoiceState = (rootState: { [VOICE_FEATURE_KEY]: VoiceState }): VoiceState => rootState[VOICE_FEATURE_KEY];
 
 export const selectAllVoice = createSelector(getVoiceState, selectAll);
+export const selectAllVoiceEntities = createSelector(getVoiceState, selectEntities);
```
huylexuan44: remove it not use

huylexuan44 (approved):

PR #7663, opened 2025-06-10
Mobile app: fix notification setting not save status mobile

hoangtranlehuy:
Mobile app: fix notification setting not save status mobile.
Issue: https://github.com/orgs/nccasia/projects/16/views/1?filterQuery=hoang&pane=issue&itemId=114080653
Expect case:
- Open notication setting in channel not setting before, check option default.
- Show correct notication type option channel.
- Save status notification type when choose other option.


hoangtranlehuy:
https://github.com/user-attachments/assets/84e0b3cf-c52c-4958-ad1b-78b4095a7b52

phuocnguyenncc (approved):

PR #7671, opened 2025-06-10
Mobile app: fix break UI when show album name mobile

hoangtranlehuy:
Mobile app: fix break UI when show album name mobile.
Issue: https://github.com/orgs/nccasia/projects/16/views/1?filterQuery=hoang&pane=issue&itemId=114377088
Expect case: render title of album with elipped size when it's too long.


hoangtranlehuy:
image

phuocnguyenncc (approved):

PR #7684, opened 2025-06-11
Fix mark As read 

sadohpv:


libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -1806,6 +1809,76 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children }) =
 		);
 	}, []);
 
+	const onMarkAsRead = useCallback(async (markAsReadEvent: MarkAsRead) => {
+		const store = getStore();
+
+		const channels = selectChannelThreads(store.getState() as RootState);
+		if (!markAsReadEvent.category_id) {
+			console.log('Mark Clan');
```
huylexuan44: clear log

huylexuan44: check null markAsReadEvent

huylexuan44 (approved):

PR #7690, opened 2025-06-12
Mobile app: fix render markdown buzz message plain text mobile

hoangtranlehuy:
Mobile app: fix render markdown buzz message plain text mobile.
Expect case: render buzz message with correct style for message buzz(red text).


hoangtranlehuy:
image

phuocnguyenncc (approved):

PR #7787, opened 2025-06-19
Mobile App: fix notification clan setting in overview miss default and nothing option

thanhlichncc:
fix: notification clan setting in overview miss default and nothing option


libs/mobile-components/src/lib/types/clanNotificationsSetting.ts
```diff
@@ -1,18 +1,19 @@
 import { NotiChannelCategorySettingEntity } from '@mezon/store-mobile';
+import { TFunction } from 'i18next';
```
phuocnguyenncc, line 2: check impact build on web

phuocnguyenncc (approved):

PR #7793, opened 2025-06-19
Mobile app: fix list suggestion member mention render wrong size mobile

hoangtranlehuy:
Mobile app: fix list suggestion member mention render wrong size mobile.
Expect case:
- When @ mention with no keyword, show here mention in last, render correct size item.
- When @ mention with keyword, show sorted list, render correct size item.


hoangtranlehuy:
https://github.com/user-attachments/assets/5d1b2bd2-c5d9-4381-8748-53cfc6c20498

phuocnguyenncc (approved):

PR #7823, opened 2025-06-23
Mobile app: fix create thread from message not send new message content mobile

hoangtranlehuy:
Mobile app: fix create thread from message not send new message content mobile.
Issue: https://github.com/orgs/nccasia/projects/16/views/1?filterQuery=hoang&pane=issue&itemId=116289353
Expect case:
- Create thread with out message, create new thread public or private with first message is sent.
- Create thread from message, new thread with show messageCreated as first message and send new message below.


hoangtranlehuy:
https://github.com/user-attachments/assets/9d9a4003-7a8d-459e-b66e-3f8988dc7868

phuocnguyenncc (approved):

PR #7829, opened 2025-06-24
Add VoiceStickers into voice call

kind6688:


libs/components/src/lib/components/VoiceChannel/MyVideoConference/Reaction/useSendVoiceSticker.ts
```diff
@@ -0,0 +1,51 @@
...
+                    type: 'voice_sticker',
+                    channel_id: currentChannel.channel_id,
+                    stickerId,
+                    stickerUrl,
+                    timestamp: Date.now()
+                };
+
+                const data = new TextEncoder().encode(JSON.stringify(eventPayload));
+
+                room.localParticipant.publishData(data, { reliable: true });
```
huylexuan44, line 30: recheck why used live kit socket ?

PR #8040, opened 2025-07-08
Mobile App: fix bug channel voice asks for permission to access the microphone it trigger pip mode

thanhlichncc:
fix bug channel voice asks for permission to access the microphone it trigger pip mode
issue: https://github.com/orgs/nccasia/projects/16/views/1?filterQuery=thanhlic&pane=issue&itemId=118808631

### DIFF
--- a/apps/mobile/src/app/screens/home/homedrawer/components/ChannelVoice/index.tsx
+++ b/apps/mobile/src/app/screens/home/homedrawer/components/ChannelVoice/index.tsx
@@ -231,6 +231,7 @@ function ChannelVoice({
 	const [isSpeakerOn, setIsSpeakerOn] = useState<boolean>(false);
 	const isPiPMode = useAppSelector((state) => selectIsPiPMode(state));
 	const dispatch = useAppDispatch();
+	const [isEnablePip, setIsEnablePip] = useState<boolean>(false);
 
 	const { sendSignalingToParticipants } = useSendSignaling();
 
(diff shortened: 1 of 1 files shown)


apps/mobile/src/app/screens/home/homedrawer/components/ChannelVoice/index.tsx
```diff
@@ -289,10 +290,14 @@ function ChannelVoice({
 
 	useEffect(() => {
 		const subscription = AppState.addEventListener('change', async (state) => {
+			if (!isEnablePip) {
+				setIsEnablePip(true);
+				return;
+			}
```
tomedis, line 296: then even the user has granted the authority, the first time it is not on mode pip

Issue only happens when request permission

apps/mobile/src/app/screens/home/homedrawer/components/ChannelVoice/index.tsx
```diff
@@ -289,10 +290,14 @@ function ChannelVoice({
...
 			if (state === 'background') {
 				if (Platform.OS === 'android') {
 					const isPipSupported = await NativeModules.PipModule.isPipSupported();
-					if (isPipSupported) {
+					if (isPipSupported && isEnablePip) {
```
phuocnguyenncc, line 300: then even the user has granted the authority, the first time it is not on mode pip

Issue only happens when request permission

phuocnguyenncc:
have fixed

PR #8043, opened 2025-07-08
fix use emoji sale on chat box

sadohpv:


libs/components/src/lib/components/EmojiPicker/index.tsx
```diff
@@ -438,6 +439,21 @@ const EmojisPanel = React.memo(function EmojisPanel({
...
+			return;
+		}
+
+		if (item.is_for_sale && item.src) {
+			const idSale = getIdSaleItemFromSource(item.src);
+			onEmojiSelect(idSale, item.shortname || '');
+			return;
+		}
+
+		onEmojiSelect(item.id || '', item.shortname || '');
```
huylexuan44: don't use || '' please check has value and use, you can add toast error info , if value empty

huylexuan44 (approved):

PR #8067, opened 2025-07-09
Mobile app: fix select permission setup role clan mobile

hoangtranlehuy:
Mobile app: fix select permission setup role clan mobile.
Expect case:
- toggle switch or item will set true status of permission.
- Press save will show loading and save all change for permission role setup.

https://github.com/user-attachments/assets/21cf6b7a-1f05-46cc-bfdb-3af653a87be2


phuocnguyenncc:
duplicate https://github.com/mezonai/mezon/pull/8058

PR #8075, opened 2025-07-10
fix bold text input

sadohpv:


libs/utils/src/lib/utils/parseHtmlAsFormattedText.ts
```diff
@@ -337,6 +337,42 @@ export const processMarkdownEntities = (text: string | undefined, entities: ApiM
 		.filter(Boolean) as IMarkdownOnMessage[];
 };
 
+export const processBoldEntities = (entities: MentionItem[], markdown: IMarkdownOnMessage[]) => {
```
huylexuan44, line 340: test more cases and record video show for me

huylexuan44 (approved):

PR #8085, opened 2025-07-11
fix upload sound and orther

kind6688:


libs/components/src/lib/components/ClanSettings/SettingSoundEffect/index.tsx
```diff
@@ -22,25 +23,37 @@ export type SoundType = {
...
 
 const SettingSoundEffect = () => {
 	const [showModal, setShowModal] = useState(false);
+	const [selectedSound, setSelectedSound] = useState<SoundType | null>(null);
 	const dispatch = useAppDispatch();
 	const currentClanId = useSelector(selectCurrentClanId) || '';
 	const currentUserId = useSelector(selectCurrentUserId) || '';
+	const currentClan = useSelector(selectCurrentClan);
+	const userProfile = useSelector(selectAllAccount);
+
```
huylexuan44: useAppselector

libs/components/src/lib/components/ClanSettings/SettingSoundEffect/index.tsx
```diff
@@ -22,25 +23,37 @@ export type SoundType = {
...
+	const sounds = useSelector((state: any) => selectAudioByClanId(state, currentClanId));
+
+	const isClanOwner = useMemo(() => {
```
huylexuan44: remove memo

huylexuan44 (approved):

PR #8107, opened 2025-07-14
fix avatar topic, avatar  emplemeral message , mention in DM action pin

van-datncc:


libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -856,7 +861,7 @@ export const sendEphemeralMessage = createAsyncThunk('messages/sendEphemeralMess
 			references,
 			false,
 			false,
-			undefined,
+			avatarToUse,
```
huylexuan44: avatarToUse || undefined

van-datncc: done

huylexuan44 (approved):

PR #8118, opened 2025-07-15
fix navigate after click

sadohpv:


libs/components/src/lib/components/ChannelSetting/Component/PermissionsChannel/listMemberPermission.tsx
```diff
@@ -15,7 +17,9 @@ const ListMemberPermission = (props: ListMemberPermissionProps) => {
 	const dispatch = useAppDispatch();
 	const rawMembers = useSelector(selectAllUserChannel(channel.channel_id || ''));
 	const currentClanId = useSelector(selectCurrentClanId);
-
+	const navigate = useNavigate();
```
huylexuan44: update to use useCustomNavigate

huylexuan44 (approved):

PR #8243, opened 2025-07-22
fix anonymous join clan

sadohpv:


libs/store/src/lib/clanMembers/clan.members.ts
```diff
@@ -122,7 +122,7 @@ export const UsersClanSlice = createSlice({
 			const { clanId, user } = action.payload;
 
 			if (!state.byClans[clanId]) {
-				state.byClans[clanId] = getInitialClanState();
```
huylexuan44: why return ?

sadohpv: If state.byClans[clanId] doesn't exist which mean api not fetch at first time so we don't need update anything

huylexuan44 (approved):

PR #8294, opened 2025-07-24
Mobile App: show displayname in account settings

thanhlichncc:
issue: https://github.com/orgs/mezonai/projects/2/views/1?filterQuery=thanhlich&pane=issue&itemId=121232224
image


apps/mobile/src/app/screens/settings/AccountSetting/index.tsx
```diff
@@ -124,6 +124,7 @@ export const AccountSetting = ({ navigation }: SettingScreenProps<AccountSetting
 			},
 			{
 				title: t('displayName'),
+				description: userProfile?.user?.display_name,
```
phuocnguyenncc: cover case undefined show empty string

phuocnguyenncc (approved):

PR #8427, opened 2025-07-30
Fix cookie consent in homepage and !isElectron

kind6688:


apps/chat/src/app/app.tsx
```diff
@@ -168,21 +168,21 @@ export function App() {
...
-		if (splashScreen) {
-			splashScreen.style.display = 'none';
-		}
-	}, []);
+    useEffect(() => {
+        const splashScreen = document.getElementById('splash-screen');
+        if (splashScreen) {
+            splashScreen.style.display = 'none';
+        }
+    }, []);
```
nguyentrannhan: check lại style

Issue #8389, opened 2025-07-29
Fix call log issue


phuocnguyenncc:

phuocnguyenncc:
CallLog {
	MISSED = 'You Missed',
	RECIPIENT_DECLINED = 'Receiver Rejected',
	YOU_DECLINED = 'You Rejected',
	VOICE_CALL = 'Audio call',
	VIDEO_CALL = 'Video call',
	CALL_BACK = 'Call back',
	INCOMING_CALL = 'Incoming call',
	OUTGOING_CALL = 'Outgoing call',
	YOU_CANCELED = 'You canceled',
	TIME_DEFAULT = '0 mins 0 secs'
}

PR #8428, opened 2025-07-30
Fix check send call log

hoangtranlehuy:
Fix check send call log.
Issue: https://github.com/mezonai/mezon/issues/8389
Expect case:
- Message call log show correct for sender and receiver.
- Show true color for some case, cancel, and out going miss call.

https://github.com/user-attachments/assets/551e4d80-65f6-43fb-80ac-2ad0d0810d33

https://github.com/user-attachments/assets/e94c104b-3d84-4465-8b79-073790875a17


libs/components/src/lib/components/CallLogMessage/CallLogMessage.tsx
```diff
@@ -88,13 +89,15 @@ export default function CallLogMessage({ userId, username, messageId, channelId,
 	const isPlayDialTone = useSelector(selectAudioDialTone);
 	const isPlayRingTone = useSelector(selectAudioRingTone);
 	const isPlayBusyTone = useSelector(selectAudioBusyTone);
-	const key = `${callLog.callLogType}_${senderId === userId ? 'SENDER' : 'RECEIVER'}`;
+	const userProfile = useSelector(selectAllAccount);
+	const isMe = useMemo(() => userProfile?.user?.id === senderId, [userProfile?.user?.id, senderId]);
```
phuocnguyenncc, line 93: Define but not used, only used for key, re-check

phuocnguyenncc (approved):

PR #8444, opened 2025-07-31
fix: checkvalidate name group

kind6688:


libs/components/src/lib/components/ChannelTopbar/index.tsx
```diff
@@ -118,11 +118,37 @@ const TopBarChannelText = memo(() => {
 	}, [currentDmGroup?.channel_label, currentDmGroup?.type, currentDmGroup?.usernames]);
 
 	const [isEditing, setIsEditing] = useState(false);
+	const [editError, setEditError] = useState<string | null>(null);
+	const [editValue, setEditValue] = useState(channelDmGroupLabel || '');
+
+	useEffect(() => {
```
huylexuan44, line 124: check duplicate validate

huylexuan44 (approved):

PR #8482, opened 2025-08-01
fix call log end call send multitime group call

hoangtranlehuy:
fix call log end call send multitime group call
Issue: https://github.com/mezonai/mezon/issues/8430
Expect case:
- Only send end call log when final participant leave.
- On mobile hide option chat and screen share when apply call from an others.

https://github.com/user-attachments/assets/a9109da6-b273-4048-9e68-95cb70e2a7f5

https://github.com/user-attachments/assets/f6533ffe-df26-4256-b49f-d5c030ae20f6

### DIFF
--- a/apps/mobile/src/app/screens/home/homedrawer/components/ChannelVoice/ControlBottomBar/ButtonEndCall.tsx
+++ b/apps/mobile/src/app/screens/home/homedrawer/components/ChannelVoice/ControlBottomBar/ButtonEndCall.tsx
@@ -49,11 +49,11 @@ const ButtonEndCall = ({ channelId, clanId, isGroupCall = false }: { channelId: 
 				dispatch(
 					messagesActions.sendMessage({
 						channelId: currentDmGroup?.channel_id,
-						clanId: '',
+						clanId: '0',
 						mode: ChannelStreamMode.STREAM_MODE_GROUP,
 						isPublic: true,
 						content: {
-							t: 'Voice call ended',
+							t: 'Group call ended',
 							callLog: {
 								isVideo: false,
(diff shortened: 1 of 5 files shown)


apps/mobile/src/app/screens/home/homedrawer/components/ChannelVoice/ControlBottomBar/ButtonEndCall.tsx
```diff
@@ -45,6 +45,29 @@ const ButtonEndCall = ({ channelId, clanId, isGroupCall = false }: { channelId:
 		};
 
 		if (type === 'quit') {
+			if (room?.numParticipants === 1) {
+				dispatch(
+					messagesActions.sendMessage({
+						channelId: currentDmGroup?.channel_id,
+						clanId: '',
```
phuocnguyenncc: clanId = 0

apps/mobile/src/app/screens/home/homedrawer/components/ChannelVoice/ControlBottomBar/ButtonEndCall.tsx
```diff
@@ -45,6 +45,29 @@ const ButtonEndCall = ({ channelId, clanId, isGroupCall = false }: { channelId:
...
+						mode: ChannelStreamMode.STREAM_MODE_GROUP,
+						isPublic: true,
+						content: {
+							t: 'Voice call ended',
```
phuocnguyenncc: change to Group call ended

phuocnguyenncc (approved):

PR #8571, opened 2025-08-05
update banner for community

kind6688:


libs/components/src/lib/components/SettingComunity/index.tsx
```diff
@@ -0,0 +1,451 @@
+"use client"
```
huylexuan44, line 1: this module not resuse, let move to chat/src/app/pages

libs/store/src/lib/comunity/comunity.slice.ts
```diff
@@ -0,0 +1,133 @@
...
+export const comunityReducer = comunitySlice.reducer;
+
+export const comunityActions = {
+    ...comunitySlice.actions,
+    updateCommunityStatus,
+    updateCommunityBanner,
+};
+
+export const selectComunityState = (state: RootState) => state[COMUNITY_FEATURE_KEY] as ComunityState;
+export const selectIsCommunityEnabled = (clanId: string) => (state: RootState) =>
```
huylexuan44: wrap create selector out side

huylexuan44: and make selector

huylexuan44 (approved):

Issue #8578, opened 2025-08-06
Group call: show in correct display name of user start group call.


hoangtranlehuy:
<img width="1440" height="824" alt="Image" src="https://github.com/user-attachments/assets/293e8767-b427-4150-9f21-36030dd813f5" />

PR #8580, opened 2025-08-06
fix call log show incorrect user display name

hoangtranlehuy:
fix call log show incorrect user display name start call.
Issue: https://github.com/mezonai/mezon/issues/8578
Expect case:
- Show display name or username of the user who start call.
- Show type of call( group, audio, video) was started.

image

https://github.com/user-attachments/assets/c28ba33b-08c7-4585-96ec-b350ba4e2f39

https://github.com/user-attachments/assets/2b0d096b-ba73-4ea4-9e41-8b9622dcb89b

### DIFF
--- a/libs/components/src/lib/components/MessageWithUser/index.tsx
+++ b/libs/components/src/lib/components/MessageWithUser/index.tsx
@@ -331,7 +331,7 @@ function MessageWithUser({
 						{!!message?.content?.callLog?.callLogType && (
 							<CallLogMessage
 								userId={userId || ''}
-								username={user?.user?.display_name || ''}
+								username={message?.display_name || message?.username || ''}
 								channelId={message?.channel_id}
 								messageId={message?.id}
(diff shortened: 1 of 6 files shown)


libs/components/src/lib/components/MessageWithUser/index.tsx
```diff
@@ -331,7 +331,7 @@ function MessageWithUser({
 						{!!message?.content?.callLog?.callLogType && (
 							<CallLogMessage
 								userId={userId || ''}
-								username={user?.user?.display_name || ''}
+								username={message?.display_name || message?.username || ''}
```
phuocnguyenncc, line 334: check impact call log in call 1:1

hoangtranlehuy: done.

phuocnguyenncc (approved):

PR #8582, opened 2025-08-06
fix: hide category when have channel

kind6688:


libs/components/src/lib/components/CreateChannelModal/CreateChannelModal.tsx
```diff
@@ -104,6 +106,10 @@ export const CreateNewChannelModal = () => {
 			navigate(channelPath);
 		}
 		clearDataAfterCreateNew();
+		if (currentClan?.clan_id) {
+			await dispatch(fetchChannels({ clanId: currentClan.clan_id, noCache: true }));
```
huylexuan44: recheck, why use no cache, you can check add new state for slice ( channel, categoreies)

huylexuan44 (approved):

PR #8597, opened 2025-08-06
fix channel mode private

sadohpv:


libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -1361,10 +1362,15 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children }) =
 				//TODO: improve update once item
 				const store = await getStoreAsync();
 				const currentChannelId = selectCurrentChannelId(store.getState() as unknown as RootState);
+				const meInClan = selectMemberClanByUserId2(store.getState(), userId as string).role_id;
```
huylexuan44: this selector use for current clan active, let recheck logic

sadohpv: I create selector by clan id

huylexuan44 (approved):

PR #8613, opened 2025-08-07
fix search thread in thread

sadohpv:


libs/store/src/lib/threads/threads.slice.ts
```diff
@@ -167,8 +167,8 @@ export const searchedThreads = createAsyncThunk('threads/searchThreads', async (
 		const mezon = await ensureSession(getMezonCtx(thunkAPI));
 		const state = thunkAPI.getState() as RootState;
 		const clanId = state?.clans?.currentClanId;
-		const channelId = state.channels?.byClans[state.clans?.currentClanId as string]?.currentChannelId;
-
+		const currentChannel = selectCurrentChannel(state);
+		const channelId = currentChannel?.parent_id && currentChannel?.parent_id !== '0' ? currentChannel?.parent_id : currentChannel?.channel_id;
```
huylexuan44: make util condition to resue logic

huylexuan44 (approved):

PR #8618, opened 2025-08-07
fix Still highlight user even though users set offline

van-datncc:


libs/components/src/lib/components/MemberProfile/MemberProfile2.tsx
```diff
@@ -42,9 +42,7 @@ export const BaseMemberProfile = ({ id, creator_id }: BaseMemberProfileProps) =>
 		showContextMenu(event, userTemplate);
 	};
 
-	const isOffline = !userMeta?.online;
-
-	
+	const isOffline = userMeta?.status === 'Invisible' ? true : !userMeta?.online;
```
huylexuan44: 'Invisible' use const to check condition

huylexuan44: ||

huylexuan44 (approved):

Issue #8663, opened 2025-08-10
Mobile App: cannot edit msg bold


phuocnguyenncc:
https://github.com/user-attachments/assets/929207d9-98d9-4520-a372-204364172fb3

PR #8679, opened 2025-08-11
Mobile app: fix edit message type bold can not fill to chatbox mobile

hoangtranlehuy:
Mobile app: fix edit message type bold can not fill to chatbox mobile.
Issue: https://github.com/mezonai/mezon/issues/8663
Expect case:
- when edit bold message show bold marker in chat box.
- when edit code message show single code marker in chat box.
- when edit codeblock message show triple code marker in chat box.

https://github.com/user-attachments/assets/34433e61-e06b-46bc-a19b-804046bd4fe2

### DIFF
--- a/libs/mobile-components/src/lib/helper/mentions/index.ts
+++ b/libs/mobile-components/src/lib/helper/mentions/index.ts
@@ -125,7 +125,6 @@ export const createFormattedString = (data: IExtendedMessage) => {
 };
 
 export const formatContentEditMessage = (message: IMessageWithUser) => {
-	console.log('log => message', message);
 	const processedContentMentionsDraft = {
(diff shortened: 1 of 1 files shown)


libs/mobile-components/src/lib/helper/mentions/index.ts
```diff
@@ -119,6 +125,7 @@ export const createFormattedString = (data: IExtendedMessage) => {
 };
 
 export const formatContentEditMessage = (message: IMessageWithUser) => {
+	console.log('log => message', message);
```
phuocnguyenncc: remove log

phuocnguyenncc (approved):

PR #8763, opened 2025-08-14
250814-DESKTOP/WEB-fix bugs related to Direct Mesaage

van-datncc:


libs/components/src/lib/components/MemberList/MemberItem.tsx
```diff
@@ -22,11 +23,15 @@ function MemberItem({ user, directMessageId, isDM = true, isMe }: MemberItemProp
 	const userCustomStatus = useAppSelector((state) => selectMemberCustomStatusById(state, user?.user?.id || '', isDM));
 	const userMetaById = useAppSelector((state) => selectDirectMemberMetaUserId(state, user?.user?.id || ''));
 	const currentUserCustomStatus = useSelector(selectAccountCustomStatus);
+	const directChannel = useAppSelector((state) => selectDirectById(state, directMessageId || ''));
```
huylexuan44: don't use direct slice for member item, update group member item to diff component

huylexuan44 (approved):

PR #8765, opened 2025-08-14
250814 - WEB/DESKTOP - remove pagination flowbite

sadohpv:


libs/components/src/lib/components/SearchMessageChannel/SearchMessageChannelRender/index.tsx
```diff
@@ -125,7 +125,7 @@ const SearchMessageChannelRender = ({ searchMessages, currentPage, totalResult,
 							</div>
 							{totalResult > 25 && (
 								<div className="mt-4 h-10">
-									<Pagination
+									{/* <Pagination
```
huylexuan44: remove don't cmt like mr Nhan require

huylexuan44 (approved):

PR #8778, opened 2025-08-14
250813-WEB-Fix: add custom macos windows action buttons to handle close windows event

Tuan2002:
- Fixed application was killed when it's windows closed.
- Keep app active in background on soft-quit action
**DEMO:**
Screen Recording 2025-08-12 at 09 19 11 (1)


apps/desktop/src/main.ts
```diff
@@ -173,6 +145,61 @@ const handleWindowAction = async (window: BrowserWindow, action: string) => {
 	}
 };
 
+const handleMacWindowsAction = async (window: BrowserWindow, action: string) => {
```
huylexuan44, line 148: move to layout

Tuan2002: I moved component to layouts, please check again.

apps/chat/src/app/layouts/MacWindowsControl.tsx
```diff
@@ -0,0 +1,102 @@
...
+import React, { useCallback, useEffect } from 'react';
+
+interface MacOSWindowControlsProps {
+	className?: string;
+}
+
+export const MacOSWindowControls: React.FC<MacOSWindowControlsProps> = ({ className }) => {
+	const dispatch = useAppDispatch();
+	const isMaximized = useAppSelector(selectIsMaximized);
+	const isWindowFocused = useAppSelector(selectIsWindowFocused);
```
huylexuan44, line 23: i think remove this selector, and use hover css to handle this case

huylexuan44: selectHoveredButton same

huylexuan44 (approved):

PR #8791, opened 2025-08-15
250815-WEB/DESKTOP-Fix/bug and enhance behavior of canvas

van-datncc:


libs/components/src/lib/components/Canvas/CanvasContent.tsx
```diff
@@ -202,12 +202,25 @@ function CanvasContent({ isLightMode, content, idCanvas, isEditAndDelCanvas, onC
 					});
 				});
 				const formats = quillRef.current?.getFormat(range) || {};
-				let nextActiveOption =
-					(formats?.header as string) || (formats?.list as string) || (formats?.blockquote === true ? 'blockquote' : 'paragraph');
-				if (nextActiveOption === 'checked' || nextActiveOption === 'unchecked') {
+				let nextActiveOption = 'paragraph';
+
+				if (formats?.header === 1) {
```
huylexuan44: make const to define object define property for per condition

huylexuan44: and break to ultil function

huylexuan44 (approved):

Issue #8759, opened 2025-08-13
[Mobile App] UI bugs of Direct Message preview


sheisThanhHien:
### Bug 1: The long message with emoji at the end of the sentence is cut off at the DM preview

<img width="1170" height="289" alt="Image" src="https://github.com/user-attachments/assets/1f360034-1747-4f71-99f5-3ae794c0b14b" />
<img width="1170" height="256" alt="Image" src="https://github.com/user-attachments/assets/79b54e14-6d1d-43e9-842d-f745b58e5188" />

sheisThanhHien:
Failed

<img width="1170" height="180" alt="Image" src="https://github.com/user-attachments/assets/4362fb07-43af-4c06-bfe7-fb0688fe90c9" />

PR #8792, opened 2025-08-15
250815-MOBILE-Fix UI direct message preview show space first mobile

hoangtranlehuy:
250815-MOBILE-Fix UI direct message preview show space first mobile
Issue: https://github.com/mezonai/mezon/issues/8759
Expect case: trim message with space character at beginning.

image
image


phuocnguyenncc (approved):

hoangtranlehuy:
Add case overlap emoji long with long text.
image
image

PR #8799, opened 2025-08-15
250815-MOBILE-Fix drop to group clan expanded not work and add animation group

thanhlichncc:
task: https://github.com/mezonai/mezon/issues/8783

https://github.com/user-attachments/assets/129364c2-86d0-4a74-a222-2b7dedf0abee


apps/mobile/src/app/components/ClanGroup/styles.ts
```diff
@@ -39,7 +39,7 @@ export const style = (themeValue: Attributes, colors: any) =>
 		groupIcon: {
 			width: size.s_42,
 			height: size.s_42,
-			borderRadius: size.s_10,
+			borderRadius: 8,
```
phuocnguyenncc: using size

phuocnguyenncc (approved):

PR #8818, opened 2025-08-18
25-08-15-DESKTOP/WEB-Fix - Trim the first line with markdown

van-datncc:


libs/utils/src/lib/utils/parseHtmlAsFormattedText.ts
```diff
@@ -160,7 +156,8 @@ function parseMarkdown(html: string) {
 
 	// Pre
 	parsedHtml = parsedHtml.replace(/`{3}([\s\S]*?)`{3}/g, function (match, p1) {
-		return '<pre>' + p1.replace(/\n/g, '___#new_line___') + '</pre>';
```
huylexuan44, line 163: test case:
content


content2

van-datncc: action done

PR #8866, opened 2025-08-19
250819-WEB/DESKTOP- Fix Message text with image is not reflected correctly with mobile

van-datncc:


van-datncc:
[#8850](https://github.com/mezonai/mezon/issues/8850) done

huylexuan44 (approved):

PR #8995, opened 2025-08-25
250825-MOBILE-fix last message DM show emoji mobile

hoangtranlehuy:
250825-MOBILE-fix last message DM show emoji mobile.
Issue: https://github.com/mezonai/mezon/issues/8759
Expect: Only show emoji which is before tail ellipsized of last message.

image
image


apps/mobile/src/app/screens/messages/DMListItemLastMessage/index.tsx
```diff
@@ -33,6 +33,32 @@ const isHeadingText = (text?: string) => {
...
+				iFormatted++;
+				iLine++;
+			} else {
+				break;
+			}
+		}
+	}
+
+	return lastIndex;
+};
```
phuocnguyenncc: logic should not be used. just using styles

phuocnguyenncc (approved):

Issue #8568, opened 2025-08-05
[Mobile] Delete Account fucntion


sheisThanhHien:
GIVEN I am a user
WHEN I go to Setting > Account 
AND I click on button Delete Account
THEN I expect to see the related data of the account should be deleted

![Image](https://github.com/user-attachments/assets/84ec6710-bb06-455f-983a-d621370959fa)

ntt-huong312:
**Steps to Reproduce:**

1. Click "Setting".
2. Click "Account".
3. Click "Delete Account".
4. Log in with the deleted account.

**Actual Result:** The deleted account data still appears.

**Expected Result:** The data should be permanently deleted after deletion.
**Evidence:** 
[https://drive.google.com/file/d/17Afv33kCnKI3MtQFNKweErNG905slOn4/view?usp=sharing](url)

PR #9050, opened 2025-08-27
250827-MOBILE-Fix bug in owner clan of del account

thanhlichncc:
issue: https://github.com/mezonai/mezon/issues/8568


libs/translations/src/languages/vi/accountSetting.json
```diff
@@ -41,7 +41,7 @@
     "toast": {
         "deleteAccount": {
             "success": "Xóa tài khoản thành công",
-            "error": "Xóa tài khoản thất bại"
```
phuocnguyenncc, line 44: deleteAccount ?? re-check impact func delete account

phuocnguyenncc (approved):

PR #9086, opened 2025-08-28
250828-DESKTOP/ fix: dont show create clan when join call in desktop

kind6688:


kind6688:
[#9087](https://github.com/mezonai/mezon/issues/9087)

huylexuan44 (approved):

Issue #9141, opened 2025-09-03
[Desktop/Website] The countdown timer event is reflected incorrectly with desktop time


sheisThanhHien:
### Steps to reproduce:
1. Create Event 
2. Set event start time 10 minutes before actual start
3. Observe the desktop time and the Countdown time event

### Expected Result: The countdown time event should reflect correctly. Example: 8 minutes left, 2 minutes left, ...

<img width="1916" height="1030" alt="Image" src="https://github.com/user-attachments/assets/612093fd-9099-431b-be41-555582001737" />

sheisThanhHien:
Case: Edit event before starting 10 minutes

<img width="695" height="456" alt="Image" src="https://github.com/user-attachments/assets/aec958f3-29dd-4927-8fd7-217ef91f6cb0" />

PR #9151, opened 2025-09-03
250903-WEB/DESKTOP-update: real time count down for event

kind6688:
[9141](https://github.com/mezonai/mezon/issues/9141)


libs/components/src/lib/components/ChannelList/EventChannelModal/ModalCreate/itemEventManagement.tsx
```diff
@@ -143,6 +143,17 @@ const ItemEventManagement = (props: ItemEventManagementProps) => {
...
 	}, [event?.event_status]);
 
+	const timeUntilEvent = useMemo(() => {
+		if (!eventIsUpcomming || !event?.start_time) return null;
+
+		const startTime = new Date(event.start_time).getTime();
+		const currentTime = new Date().getTime();
+		const diffInMs = startTime - currentTime;
+		const diffInMinutes = Math.ceil(diffInMs / (1000 * 60));
+		if (diffInMinutes === 1) return '1 minute left. Join in!';
```
huylexuan44: when time less than 1 ?

kind6688: only count down by minute , sir

huylexuan44: add support vi and en for this msg , add message for json language file

Issue #9168, opened 2025-09-04
Popup mention dont close when send message or click outside


kind6688:
<img width="724" height="551" alt="Image" src="https://github.com/user-attachments/assets/f3fddd09-5770-442e-af6c-ac5507b8b4ca" />

PR #9169, opened 2025-09-04
250904-WEB/DESKTOP - fix: click outside for close popup mention

kind6688:
[9168](https://github.com/mezonai/mezon/issues/9168)


huylexuan44:
I will recheck

Issue #8839, opened 2025-08-18
[Mobile App] Issue on group clan when switch account


phuocnguyenncc:
<img width="337" height="540" alt="Image" src="https://github.com/user-attachments/assets/1dafbb27-fbc6-44f8-9c6f-0ff1059a830f" />

sheisThanhHien:
### FAILED. 
### Expected Result: group clan ở acc A --> qua acc B --> chuyển lại acc A thì ER group đó vẫn i như cũ
### Actual Result: group clan ở acc A bị mất
https://github.com/user-attachments/assets/1cdb5034-492f-4ad2-b57e-9e7d7cd0fd57

ntt-huong312:
**FAILED on version 1.2.65 (15)**

**Expected Result:** group clan ở acc A --> qua acc B --> chuyển lại acc A thì ER group đó vẫn i như cũ
 
https://drive.google.com/file/d/1fw8DYxJHZot4L2zi2dAs_0PQnpy76BBs/view?usp=drive_link

PR #9429, opened 2025-09-12
250912-MOBILE-fix order clan group not backup after switch account

thanhlichncc:
issue: https://github.com/mezonai/mezon/issues/8839


phuocnguyenncc:
will do later, avoid Impact

Issue #9466, opened 2025-09-15
[Desktop/Website] Delete Category bugs


sheisThanhHien:
## Bug 1: Delete category contains channel/thread
**Steps to reproduce:** 
1. Create a category
2. Create channel/ thread, voice channel, stream channel in category 
3. Right click on Category --> Edit Category
4. Delete Category

**Actual Result:** The channel/thread, voice/stream channel is pushed up, not disappear immediately
**Expected Result:** All of Category's channel/thread should be deleted immediately, Redirect to Welcome Channel

https://github.com/user-attachments/assets/08ef2db5-1627-4899-bda2-7012dbca8104

## Bug 2: Delete Category does not contain any channel/thread at all
**Steps to reproduce:** 
1. Create a category
2. Right click on Category --> Delete Category

**Actual Result:** Display empty channel, Redirect to Welcome Channel 
**Expected Result:**  Redirect to Welcome Channel

https://github.com/user-attachments/assets/a040121b-01d3-469a-b384-83d4d62b097b

PR #9520, opened 2025-09-17
250917- WEB/DESKTOP - fix: delete clan dont redirect and kickout member

kind6688:


kind6688:
[#9466](https://github.com/mezonai/mezon/issues/9466)

libs/core/src/lib/chat/hooks/useCategory.ts
```diff
@@ -1,30 +1,108 @@
...
+					clan_id: channel.clan_id || '',
+					deletor: 'user',
+					channel_type: channel.type || 0,
+					channel_label: channel.channel_label || '',
+					category_id: channel.category_id || '',
+					parent_id: channel.parent_id || ''
+				};
+				dispatch(channelsActions.deleteChannelSocket(channelDeleteEvent));
+				dispatch(listChannelsByUserActions.remove(channel.id));
+				dispatch(listChannelRenderAction.updateClanBadgeRender({ channelId: channel.id, clanId: channel.clan_id || '' }));
```
huylexuan44: make bulk remove action instead  use like this

libs/core/src/lib/chat/hooks/useCategory.ts
```diff
@@ -1,30 +1,108 @@
...
 
+	const navigateAfterDeleteCategory = (categoryId: string) => {
+		const store = getStore();
+		const state = store.getState();
+		const allChannels = selectAllChannels(state);
+		const currentChannel = currentChannelId ? allChannels.find((ch) => ch.id === currentChannelId) : null;
+		if (!currentChannel || currentChannel.category_id !== categoryId) {
+			return;
+		}
+		const remainingChannels = allChannels.filter((ch) => ch.category_id !== categoryId && !checkIsThread(ch));
```
huylexuan44: I think we need navigate to wellcome channel  for this case

huylexuan44 (approved):

PR #9658, opened 2025-09-23
250923 - WEB/DESKTOP - fix check pip

sadohpv:


libs/store/src/lib/threads/threads.slice.ts
```diff
@@ -100,6 +100,7 @@ export const fetchThreadsCached = async (
 			time: channelData.cache?.lastFetched || Date.now()
 		};
 	}
```
huylexuan44, line 102: clear log

huybuidoanquang (approved):

Issue #9784, opened 2025-09-30
[Desktop/Website] UI bug when creating thread with long message


sheisThanhHien:
<img width="2880" height="1704" alt="Image" src="https://github.com/user-attachments/assets/b23bae08-e131-4639-9927-67e8c94b8734" />

Issue #9657, opened 2025-09-23
[Desktop/Website] Lost displayed init message when creating thread


sheisThanhHien:
**Steps to reproduce:**
1. At channel, send a message
2. Right-Click on that message --> Create Thread
3. Create thread window opens, enter thread name under 3 characters and enter Starter Message
4. Press Enter and Observe

**Actual Result:** Lost displayed init message when enter thread name under 3 characters

https://github.com/user-attachments/assets/08e3551c-5db5-40c9-a3ce-cbbe5927271b

PR #9822, opened 2025-10-01
251001 - WEB /DESKTOP fix: Bug create thread dont keep message when show error and bug UI with message so long 

kind6688:


kind6688:
[#9784](https://github.com/mezonai/mezon/issues/9784)
[#9657](https://github.com/mezonai/mezon/issues/9657)

huylexuan44 (approved):

Issue #9515, opened 2025-09-17
[Desktop/Website] UI bug Reply


sheisThanhHien:
<img width="832" height="345" alt="Image" src="https://github.com/user-attachments/assets/0681db25-d094-4980-888a-ec6151a167c1" />

PR #9938, opened 2025-10-08
251008-WEB/DESKTOP / fix: UI reply , esc Modal, text modal delete thread,channel

kind6688:
https://github.com/mezonai/mezon/issues/9515


kind6688:
image -> reply 
image

image -> 
image


can't esc 
image
image
...

image
- >

libs/core/src/lib/chat/hooks/useEscapeKeyClose.ts
```diff
@@ -1,25 +1,32 @@
...
+					onClose();
+				}
 			}
 		};
 
+		if (element.tabIndex < 0) {
+			element.tabIndex = -1;
+		}
+
+		document.addEventListener('keydown', handleKeyDown);
```
huylexuan44: don't use like this, check old modal to handle

huylexuan44 (approved):

kind6688:
[#9515](https://github.com/mezonai/mezon/issues/9515)

Issue #9973, opened 2025-10-09
[Desktop/Website] Remove quick menu for Stream channel


sheisThanhHien:
<img width="1488" height="521" alt="Image" src="https://github.com/user-attachments/assets/a2b27865-2465-4c4b-b5b4-8616f23d35ee" />

PR #9980, opened 2025-10-09
251009- WEB/DESKTOP - remove qick menu in channel stream

kind6688:


kind6688:
[#9973](https://github.com/mezonai/mezon/issues/9973)

Issue #10057, opened 2025-10-14
[Desktop/Website] Check condition transferring token


sheisThanhHien:
<img width="583" height="619" alt="Image" src="https://github.com/user-attachments/assets/10fec773-c4fa-4b65-aa8f-efd8fe2f996a" />

sheisThanhHien:
<img width="523" height="616" alt="Image" src="https://github.com/user-attachments/assets/c474c49a-3494-4c61-a9f8-bd174bdcf2a6" />

PR #10091, opened 2025-10-15
251015 WEB/DESKTOP - limit send token

kind6688:
image
[#10057](https://github.com/mezonai/mezon/issues/10057)


libs/components/src/lib/components/ModalUserProfile/StatusProfile/ModalSendToken.tsx
```diff
@@ -66,6 +69,16 @@ const ModalSendToken = ({
 		};
 	}, [setToken]);
 
+	useEffect(() => {
+		setWalletBalanceError(null);
+	}, [walletDetail?.balance]);
+
+	useEffect(() => {
+		if (token <= 99000000) {
```
phuocnguyenncc: define constant for: 99000000 100000000 100000099

kind6688: done sir

Issue #10106, opened 2025-10-16
[Desktop/Website] Display username in topic box in Clan


sheisThanhHien:
**Env:** mezon dev

**Steps to reproduce:**
1. Click on create topic on context menu's message 
2. Send stater message to create topic on topic box  
3. Refresh app

<img width="1583" height="225" alt="Image" src="https://github.com/user-attachments/assets/cc1e4248-cbc4-4b4b-a992-391bbceadadd" />

PR #10124, opened 2025-10-16
251016 WEB/DESKTOP - fix:name user in topic and logic jump 

kind6688:
[#10106](https://github.com/mezonai/mezon/issues/10106)

### DIFF
--- a/libs/components/src/lib/hooks/useNotificationJump.ts
+++ b/libs/components/src/lib/hooks/useNotificationJump.ts
@@ -32,74 +32,78 @@ export const useNotificationJump = ({ messageId, channelId, clanId, topicId, isT
 	const handleJumpToTopic = useCallback(async () => {
 		if (!topicId || !channelId || !clanId) return;
 
-		const currentState = getStore().getState();
-		const currentChannelId = selectCurrentChannelId(currentState);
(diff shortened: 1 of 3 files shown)


libs/components/src/lib/hooks/useNotificationJump.ts
```diff
@@ -26,29 +28,79 @@ interface UseNotificationJumpProps {
...
-	const currentChannelId = useAppSelector(selectCurrentChannelId);
 	const isShowCanvas = useSelector(selectIsShowCanvas);
-
 	const handleJumpToTopic = useCallback(async () => {
-		if (!topicId || !channelId || !clanId || !messageId) return;
+		if (!topicId || !channelId || !clanId) return;
 
+		const currentState = getStore().getState();
+		const currentChannelId = selectCurrentChannelId(currentState);
+		const currentClanId = selectCurrentClanId(currentState);
```
phuocnguyenncc: a lot of logic, should wrap func handleJumpToTopic in a try-catch

kind6688: yes sir

phuocnguyenncc (approved):

Issue #10485, opened 2025-11-04
[Mobile App] Clan nickname does not update after removing


ntt-huong312:
**Steps to Reproduce:**

1. Tap on Setting Profile
2. Tap on Clan
3. Tap the **x** icon to remove the current clan name
4. Tap on Save

**Actual Result:** The previous clan name still remains in the profile

**Expected Result:** 
- If clanNickname is not available --> display displayName
- If displayName is not available --> display username

https://github.com/user-attachments/assets/e78bfdf8-4363-4ba7-a01d-b8f68c12da88

ntt-huong312:
### Cannot remove clan nickname 

https://github.com/user-attachments/assets/a49d79b5-5d3d-4981-a396-26930b08a8d3

PR #10494, opened 2025-11-04
251104-MOBILE-Fix clan nickname not update after remove

thanhlichncc:
issue: https://github.com/mezonai/mezon/issues/10485


libs/store/src/lib/clanProfile/clanProfile.slice.ts
```diff
@@ -78,38 +78,35 @@ export const updateUserClanProfile = createAsyncThunk(
...
-					})
-				);
-
-				if (avatarUrl && currentUser?.user?.id) {
-					setUserClanAvatarOverride(currentUser.user.id, clanId, avatarUrl);
-					thunkAPI.dispatch(accountActions.incrementAvatarVersion());
-				}
-
-				thunkAPI.dispatch(fetchUserClanProfile({ clanId }));
+			const response = await mezon.client.updateUserProfileByClan(mezon.session, clanId, body as ApiUpdateClanProfileRequest);
```
phuocnguyenncc, line 88: fix at the point where use updateUserClanProfile function

phuocnguyenncc:
fixed on: https://github.com/mezonai/mezon/pull/10528

PR #10557, opened 2025-11-07
251107 - WEB/DESKTOP - fix constant combined

sadohpv:


libs/utils/src/lib/constant/index.ts
```diff
@@ -1,6 +1,6 @@
 import type { ThreadError } from '../types';
 
-export const TIME_COMBINE = 120;
+export const TIME_COMBINE = 10 * 60 * 1000; // 10 minutes in milliseconds
```
nguyentrannhan: TIME_COMBINE_MS

huylexuan44 (approved):

Issue #10811, opened 2025-11-21
[Desktop/Website] Username is not displayed in User Profile after opening new DM


sheisThanhHien:
<img width="1568" height="773" alt="Image" src="https://github.com/user-attachments/assets/4653db6d-945f-418d-bcb6-cbb78f2763e5" />

sheisThanhHien:
## Additional cases: Register new account --> Self chat

<img width="1719" height="948" alt="Image" src="https://github.com/user-attachments/assets/b5fb94d8-e60f-4393-848c-49c2f6a872ba" />

van-datncc:
<img width="1284" height="605" alt="Image" src="https://github.com/user-attachments/assets/495dab8a-1ec6-40c7-bcf3-ddd358be89ba" />

PR #10849, opened 2025-11-24
251124-WEB/DESKTOP-fix(dm): display username in profile view after starting new DM

van-datncc:
Task : [Desktop/Website] Username is not displayed in User Profile after opening new DM
[#10811](https://github.com/mezonai/mezon/issues/10811)
Demo : 

https://github.com/user-attachments/assets/e77d6ed3-c0d6-46dc-aee7-5bbe9f0e012b

add case my profile new account 
image

### DIFF
--- a/libs/core/src/lib/chat/hooks/useUserById.ts
+++ b/libs/core/src/lib/chat/hooks/useUserById.ts
@@ -39,7 +39,7 @@ export const useUserMetaById = (userID: string | undefined): any | undefined => 
 	});
 };
 
-export const useUserByUserId = (userID: string | undefined): ChannelMembersEntity | any => {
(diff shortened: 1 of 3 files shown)


libs/core/src/lib/chat/hooks/useUserById.ts
```diff
@@ -39,7 +39,7 @@ export const useUserMetaById = (userID: string | undefined): any | undefined =>
 	});
 };
 
-export const useUserByUserId = (userID: string | undefined): ChannelMembersEntity | undefined => {
+export const useUserByUserId = (userID: string | undefined): ChannelMembersEntity | any => {
```
huylexuan44: define type

Issue #10845, opened 2025-11-24
[Desktop/Website] Channels section reflects incorrectly after creating channel/thread


sheisThanhHien:
**Steps to reproduce:**
1. Create new channel/thread 
2. Click on channels section and observe

**Actual Result:** Channels section does not update

<img width="1317" height="900" alt="Image" src="https://github.com/user-attachments/assets/f1f1a0be-d023-4507-bc46-b84deae392fb" />

sheisThanhHien:
## TESTING ON DEV - 28/11/2025
### 1/ Update channel name reflects incorrect icon
https://github.com/user-attachments/assets/3c74066c-3469-409a-b802-e3c52108145a

### 2/ Update thread name does not update in Channels section
https://github.com/user-attachments/assets/042afc66-1d63-4072-9169-6c33b119ecf6

PR #10876, opened 2025-11-25
251125 WEB/DESKTOP - fix : list channel , dont have socket

kind6688:
[#10845](https://github.com/mezonai/mezon/issues/10845)

### DIFF
--- a/libs/store/src/lib/clans/clanSettingChannel.slice.ts
+++ b/libs/store/src/lib/clans/clanSettingChannel.slice.ts
@@ -148,61 +148,57 @@ export const settingClanChannelSlice = createSlice({
 	reducers: {
 		addChannelFromSocket: (state, action) => {
 			const channel = action.payload;
-			if (channel?.id) {
-				if (channel.parent_id && channel.parent_id !== '0') {
-					state.threadCount += 1;
-					if (!state.threadsByChannel[channel.parent_id]) {
-						state.threadsByChannel[channel.parent_id] = [];
-					}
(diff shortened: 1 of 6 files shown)


libs/store/src/lib/clans/clanSettingChannel.slice.ts
```diff
@@ -145,7 +145,68 @@ export const fetchChannelSettingInClan = createAsyncThunk(
...
+						state.threadsByChannel[parentId] = threads.filter((t) => t.id !== channelId);
+						state.threadCount = Math.max(0, state.threadCount - 1);
+						break;
+					}
+				}
+			}
+		},
+		updateChannelFromSocket: (state, action) => {
+			const channel = action.payload;
+			if (channel?.id) {
```
huylexuan44: clean your code return early and using hof some function to find parent channel with payload channel parrent id , then set new state to store

libs/store/src/lib/clans/clanSettingChannel.slice.ts
```diff
@@ -145,7 +145,68 @@ export const fetchChannelSettingInClan = createAsyncThunk(
...
+					if (!existingThread) {
+						state.threadsByChannel[channel.parent_id].push(channel);
+					}
+				} else {
+					channelSettingAdapter.addOne(state, channel);
+					state.channelCount += 1;
+				}
+			}
+		},
+		removeChannelFromSocket: (state, action) => {
```
huylexuan44, line 166: check cmt bottom updateChannelFromSocket

Issue #10862, opened 2025-11-25
[Mobile App] The user can send messages after changing a channel from public to private


ntt-huong312:
**Steps to Reproduce:**
1. User A is an Owner Clan set any channel public to private
2. In side user B --> open that priavate channel
3. User B sends a message

**Actual Result:**
- At step 2-3, After changing a channel from Public to Private, the permission update is not applied immediately can still view the channel and send messages.

**Expected Result:** Users without permission should immediately lose access and should not be able to send messages once the channel becomes Private.

ntt-huong312:
### Failed: User can stills send message after changing a channel from public to private

https://github.com/user-attachments/assets/f1c42dc6-a92e-4f95-af1a-bf832453c0e5

PR #10882, opened 2025-11-25
251125-MOBILE-Fix change permission channel not close chat screen mobile

hoangtranlehuy:
251125-MOBILE-Fix change permission channel not close chat screen mobile
Issue: https://github.com/mezonai/mezon/issues/10862
Change: Emit event to close chat when channel change permission.

### DIFF
--- a/libs/core/src/lib/chat/contexts/ChatContext.tsx
+++ b/libs/core/src/lib/chat/contexts/ChatContext.tsx
@@ -795,7 +795,7 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
 							channelType: user.channel_type
 						});
 					}
-					if (channelId === user.channel_id) {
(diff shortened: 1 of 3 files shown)


libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -872,6 +883,24 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children }) =
...
+										clanId: clanIdToJump
+									})
+								);
+							}
+							MobileEventEmitter.emit('@ON_REMOVE_USER_CHANNEL', {
+								channelId: '',
+								channelType: 0,
+								isRemoveClan: true
+							});
+						}
```
phuocnguyenncc, line 905: return

phuocnguyenncc (approved):

Issue #10758, opened 2025-11-19
[Desktop/Website] Emojis remain visible after deleting clan


sheisThanhHien:
### Xóa clan nhưng emoji vẫn còn hiện trong panel

<img width="567" height="540" alt="Image" src="https://github.com/user-attachments/assets/442a684c-483b-4923-946f-7aba72924812" />

PR #10913, opened 2025-11-26
25-11-26 WEB / DESKTOP fix : delete clan emoji, sticker, sound when leave clan

kind6688:
[#10758](https://github.com/mezonai/mezon/issues/10758)


libs/store/src/lib/emojiSuggestion/emojiSuggestion.slice.ts
```diff
@@ -95,8 +95,21 @@ export const fetchEmoji = createAsyncThunk(
 			if (!response?.emoji_list) {
 				throw new Error('Emoji list is undefined or null');
 			}
+
+			const state = thunkAPI.getState() as RootState;
+			const userClans = state.clans?.entities || {};
```
huylexuan44, line 100: check variable name?
and why need add this logic, filter in client ?

Issue #10785, opened 2025-11-20
[Mobile app] Optimize get last send user display name direct message item


hoangtranlehuy:

ntt-huong312:
**Root casuse:** The sender name does not displayed when the 3rd member send the message

ntt-huong312:
**The sender name is displayed incorrectly**
**Steps to Reproduce:**
1. User A in group DM chat -> Send a message
2. Back to DM list
3. User B send a message to DM group
4. Observe the sender name in screen user A

**Actual Result:** The sender name of user A is displayed instead of user B
**Expected Result:** Should be displayed correctly by the sender

https://github.com/user-attachments/assets/d95ec312-f9a9-4ec6-8e9a-0f3788e27ad4

PR #10934, opened 2025-11-27
251127-MOBILE-Fix last send user display name incorrect mobile

hoangtranlehuy:
251127-MOBILE-Fix last send user display name incorrect mobile
Issue: https://github.com/mezonai/mezon/issues/10785
Change: check type of direct message by message mode.

https://github.com/user-attachments/assets/1f8581c0-804e-4f70-94a6-6d94123d98b0

### DIFF
--- a/libs/store/src/lib/direct/direct.slice.ts
+++ b/libs/store/src/lib/direct/direct.slice.ts
@@ -721,7 +721,7 @@ export const directSlice = createSlice({
 		},
 		updateMoreData: (state, action: PayloadAction<DirectEntity>) => {
 			const data = action.payload;
-			const channelId = data.id;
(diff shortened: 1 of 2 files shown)


libs/store/src/lib/direct/direct.slice.ts
```diff
@@ -721,61 +721,39 @@ export const directSlice = createSlice({
...
-							user_ids: data?.user_ids
-						};
-					}
-				} else {
-					changes = {
-						...data,
-						...currentData
-					};
-				}
+			const channelId = data.id;
```
phuocnguyenncc: check data.channel_id or data.id

phuocnguyenncc (approved):

Issue #10832, opened 2025-11-24
[Mobile App] Status duration reflect does not incorrectly after creation


ntt-huong312:
**Steps to Reproduce:**
1. Tap the set custom status
2. Select any status duration 
3. Tap the Save button
4. Open set custom status again
5. Observe the status after the change

**Actual Result:** The status duration does not change after saving
**Expected Result:** The status duration should be updated correctly after changing

https://github.com/user-attachments/assets/ebb9fe25-0596-4221-8b9c-3680758f6600

PR #10957, opened 2025-11-30
251130-MOBILE-Fix custom status not sync

thanhlichncc:
issue: https://github.com/mezonai/mezon/issues/10832


apps/mobile/src/app/screens/profile/ProfileScreen.tsx
```diff
@@ -122,14 +122,14 @@ const ProfileScreen = ({ navigation }: { navigation: any }) => {
 		const data = {
 			children: (
 				<AddStatusUserModal
-					userCustomStatus={userMemberStatus?.status || ''}
+					userCustomStatus={userProfile?.user?.user_status || ''}
```
phuocnguyenncc, line 125: get from userCustomStatus = useSelector(selectAccountCustomStatus);

phuocnguyenncc:
fixed

Issue #11007, opened 2025-12-03
[Desktop/Website] User profile in DM displays incorrectly by inviting new user to clan


sheisThanhHien:
**Steps to reproduce:** 
1. Add friends to the new account
2. Go to any clan --> Invite people
3. Scroll to the bottom of the list --> Click invite on that user's entry
4. Go to DM
5. Open DM with that user and observe

<img width="1331" height="836" alt="Image" src="https://github.com/user-attachments/assets/30bd91b3-6e2d-4869-8a2d-e16f3687e7b6" />

PR #11041, opened 2025-12-04
251204-WEB/DESKTOP-fix: new user profile dm  when invite clan

van-datncc:
Task :  [Desktop/Website] User profile in DM displays incorrectly by inviting new user to clan
[#11007](https://github.com/mezonai/mezon/issues/11007)

DEMO : 

https://github.com/user-attachments/assets/e75f73f7-697b-4774-b5d4-47a2de1b69af

### DIFF
--- a/libs/components/src/lib/components/ModalUserProfile/index.tsx
+++ b/libs/components/src/lib/components/ModalUserProfile/index.tsx
@@ -71,6 +71,7 @@ const ModalUserProfile = ({
 	classWrapper,
 	classBanner,
 	hiddenRole,
+	showNote,
 	message,
 	showPopupLeft,
 	mode,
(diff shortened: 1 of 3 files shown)


libs/components/src/lib/components/ModalUserProfile/index.tsx
```diff
@@ -112,7 +110,22 @@ const ModalUserProfile = ({
 	const date = new Date(userById?.user?.create_time as string | Date);
 	const { timeFormatted } = useFormatDate({ date });
 
-	const avatarByUserId = isDM ? userById?.user?.avatar_url : userById?.clan_avatar || userById?.user?.avatar_url;
+	const effectiveUserData = useMemo(() => {
```
huylexuan44: recheck: currently DM list has full data

huylexuan44: why you need get friend state ?

huylexuan44 (approved):

Issue #11006, opened 2025-12-03
[Desktop/Website] Add optional message input field and re-design preview forwarded message


sheisThanhHien:
## Req 1:  Add input field "Add an optional message ..." into Forward Message 

<img width="648" height="770" alt="Image" src="https://github.com/user-attachments/assets/b32892b9-bf1f-4c14-bf35-241c81b3d655" />

### The optional message should be displayed at the bottom of the forwarded message ==>
<img width="580" height="499" alt="Image" src="https://github.com/user-attachments/assets/737015cd-6370-4cb9-ac5d-d6853d67eb4b" />

#### Cases need to be handled:
1. Text

<img width="1170" height="276" alt="Image" src="https://github.com/user-attachments/assets/a80d0f12-0440-4cd6-b2e0-e9bfeaa93df7" />

3. Text with image 

<img width="1160" height="270" alt="Image" src="https://github.com/user-attachments/assets/af6aa384-a084-4171-adb2-fa20ea591163" />

4. Text with video

<img width="1164" height="242" alt="Image" src="https://github.com/user-attachments/assets/d15e28c9-03ae-4f2c-af7e-ddb6f6c59569" />

5. Image

<img width="1170" height="241" alt="Image" src="https://github.com/user-attachments/assets/f28aacb2-a588-49a8-acda-a8f727257212" />

6. Video

<img width="1170" height="276" alt="Image" src="https://github.com/user-attachments/assets/1098078c-a36a-4c6e-b5c6-616d75caa9ca" />

7. Markdown 

<img width="1168" height="180" alt="Image" src="https://github.com/user-attachments/assets/e0a6fd23-0c06-4eb2-a525-cd36cdf8d045" />

8. Emoji 

<img width="323" height="91" alt="Image" src="https://github.com/user-attachments/assets/5c81c2c3-c4f6-4c32-bcf5-f92964713260" />

10. Sticker
11. Sound Effect
12. Files

<img width="1157" height="241" alt="Image" src="https://github.com/user-attachments/assets/47729c73-a384-4c9d-95c5-22571099d9df" />

sheisThanhHien:
## Failed case: Forward All Message

Nếu Forward nhiều message (trong vòng 10p) mà chứa nhiều media thì nên hiển thị như sau:

<img width="1162" height="247" alt="Image" src="https://github.com/user-attachments/assets/57006063-5900-481f-a545-f396385932d4" />

PR #11082, opened 2025-12-08
251208 - WEB /DESKTOP - update foward with attachment

kind6688:
[#11006](https://github.com/mezonai/mezon/issues/11006)

### DIFF
--- a/libs/components/src/lib/components/ForwardMessage/index.tsx
+++ b/libs/components/src/lib/components/ForwardMessage/index.tsx
@@ -20,6 +20,7 @@ import {
 	useAppDispatch,
 	useAppSelector
 } from '@mezon/store';
+import { Icons } from '@mezon/ui';
 import type { ChannelThreads, UsersClanEntity } from '@mezon/utils';
 import {
 	FOR_1_HOUR_SEC,
@@ -27,6 +28,9 @@ import {
 	TypeSearch,
 	addAttributesSearchList,
 	getAvatarForPrioritize,
+	isFileAttachment,
+	isImageFileType,
+	isVideoFileType,
 	normalizeString,
 	removeDuplicatesById
 } from '@mezon/utils';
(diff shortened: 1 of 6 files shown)


libs/components/src/lib/components/ForwardMessage/index.tsx
```diff
@@ -66,6 +70,10 @@ const ForwardMessageModal = () => {
 	const isForwardAll = useSelector(getIsFowardAll);
 	const [selectedObjectIdSends, setSelectedObjectIdSends] = useState<ObjectSend[]>([]);
 	const [searchText, setSearchText] = useState('');
+	const [forwardMessage, setForwardMessage] = useState('');
+
+	const remainingChars = 2000 - forwardMessage.length;
```
huylexuan44: reuse const

libs/core/src/lib/chat/hooks/useSendForwardMessage.ts
```diff
@@ -49,6 +49,19 @@ export function useSendForwardMessage() {
 					message.attachments
 				);
 
+				if (additionalMessage && additionalMessage.trim()) {
+					const trimmedMessage = additionalMessage.trim();
+
+					if (trimmedMessage.length > 2000) {
```
huylexuan44: reuse const

libs/components/src/lib/components/ForwardMessage/index.tsx
```diff
@@ -92,7 +100,27 @@ const ForwardMessageModal = () => {
 		return isForwardAll ? handleForwardAllMessage() : sentToMessage();
 	};
 
+	const sendMultipleMessages = async (combineMessages: MessagesEntity[], clanId: string, channelId: string, mode: number, isPublic: boolean) => {
```
huylexuan44, line 106: What if we choose to send to 100 people? We should add a status showing how many people are currently being sent; otherwise, the UI will look frozen.

PR #11185, opened 2025-12-15
251115 - WEB/DESKTOP - fix select default embed

sadohpv:
### Changes

- Add data default embed select


libs/components/src/lib/components/MessageActionsPanel/components/MessageSelect.tsx
```diff
@@ -167,6 +167,22 @@ export const MessageSelect: React.FC<MessageSelectProps> = ({ select, messageId,
 		return <>{menuItems}</>;
 	}, [availableOptions]);
 
+	useEffect(() => {
+		if (select?.valueSelected) {
```
huylexuan44: return early

sadohpv: Done

huylexuan44 (approved):

Issue #11184, opened 2025-12-14
[Website] Join clan pop-up stills remains visile after joining clan successfully


sheisThanhHien:
## Đã join clan rồi nhưng pop-up Join clan vẫn hiển thị

<img width="1424" height="842" alt="Image" src="https://github.com/user-attachments/assets/ad4929c2-ef43-482a-a04c-bd786c3a6049" />

PR #11189, opened 2025-12-15
251215/web/desktop-commit fix first-join

Luc1404:


Luc1404:
https://github.com/mezonai/mezon/issues/11184

Issue #11186, opened 2025-12-15
[Desktop/Website] Adjust UI context menu in voice channel


sheisThanhHien:
<img width="479" height="214" alt="Image" src="https://github.com/user-attachments/assets/c163d157-75e2-40a5-96f6-a2769211c71a" />

PR #11192, opened 2025-12-15
251215 - WEB/DESKTOP - fix: contextMenu in voice

kind6688:
[#11186](https://github.com/mezonai/mezon/issues/11186)


libs/components/src/lib/components/VoiceChannel/MyVideoConference/VoiceContextMenu/VoiceContextMenu.tsx
```diff
@@ -0,0 +1,210 @@
...
+			}
+			// No valid audio publication found
+			setIsMicOn(false);
+		};
+
+		// Initial check - no timeout needed as we listen to all events
+		updateMicStatus();
+
+		// Listen to track events
+		participant.on('trackMuted', updateMicStatus);
```
huylexuan44, line 72: I can't find where the event trigger is coming from. ?

PR #11217, opened 2025-12-16
251216-WEB/DESKTOP-Fix(topic): cannot use anonymous in topic

van-datncc:
Task : [Desktop/Website] Cannot use anonymous in topic
[#11205](https://github.com/mezonai/mezon/issues/11205)

Demo : 

https://github.com/user-attachments/assets/bdf9eb3c-130e-47b7-ae8b-c420ab09a142

### DIFF
--- a/apps/chat/src/app/pages/main/index.tsx
+++ b/apps/chat/src/app/pages/main/index.tsx
@@ -29,6 +29,7 @@ import {
 	e2eeActions,
 	fetchDirectMessage,
 	getIsShowPopupForward,
+	getStore,
 	onboardingActions,
 	selectAllAppChannelsListShowOnPopUp,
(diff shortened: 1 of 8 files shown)


apps/chat/src/app/pages/main/index.tsx
```diff
@@ -105,6 +107,8 @@ function MyApp() {
 
 	const { currentURL, directId } = useAppParams();
 	const memberPath = `/chat/clans/${currentClanId}/member-safety`;
+	const isFocusTopicBox = useSelector(selectClickedOnTopicStatus);
```
huylexuan44: getStore() instead of use selector

huylexuan44 (approved):

Issue #11390, opened 2025-12-26
[Mobile] UI bugs v7


ntt-huong312:
**Default Avatar**

<img width="1055" height="1280" alt="Image" src="https://github.com/user-attachments/assets/16831894-75ca-4fb0-9fb8-c4198597e317" />

<img width="548" height="1280" alt="Image" src="https://github.com/user-attachments/assets/1492b769-fdbb-45d3-bda2-425bbdad2c86" />

<img width="592" height="1280" alt="Image" src="https://github.com/user-attachments/assets/2e57b976-0809-4ef5-8def-65b4f165ca40" />

ntt-huong312:
**1/ In share screen:**

<img width="647" height="521" alt="Image" src="https://github.com/user-attachments/assets/24396962-4f0f-4841-87ae-a7c67b139a25" />


**Created event:**

<img width="592" height="1280" alt="Image" src="https://github.com/user-attachments/assets/eb856ab1-c51b-463a-a3d6-a8405c7c5ce6" />

**2/ Search result list is not cleared after removing the selected user**

**Actual Result**
The previous member search result list is still displayed after the selected user is removed from the input.

**Expected Result**
The member search result list should be cleared after the selected user is removed from the input.

https://github.com/user-attachments/assets/1682ed27-51c6-473b-b82d-b29271ac4db9

PR #11652, opened 2026-01-14
260114-MOBILE-Fix bug UI v7

thanhlichncc:
issue: https://github.com/mezonai/mezon/issues/11390


phuocnguyenncc:
@thanhlichncc create PR to develope

Issue #11868, opened 2026-01-26
[Desktop/Website] Configure voice settings


sheisThanhHien:
### 1. User Story Statement
> As a **Mezon user**, I want to **easily set up and test my audio devices (microphone and speaker), adjust volumes, and run a mic test**, so that I can **communicate clearly** with friends and teammates, ensuring they hear me well and I hear them properly.

---
#### Microphone
- User can select the input device (microphone) from available options
- User can adjust Microphone Volume to control sensitivity

#### Speaker (Output Device)
- User can select the output device (speaker/headphones) from available options
- User can adjust Speaker Volume to control playback loudness

#### Mic Test
- A Mic Test button allows the user to record and playback their voice
- User sees visual feedback (e.g., input level bar) when speaking into the microphone

<img width="1270" height="493" alt="Image" src="https://github.com/user-attachments/assets/c3109724-b0cb-452e-a2bd-f2b26ca59a9d" />

---

### 3. Scenario
Given I open Mezon on mobile or desktop
When I go to **Settings → App Settings → section Voice**
Then 
- I select my USB microphone as the Input Device (Microphone)
- I select my Bluetooth headphones as the Output Device (Speaker)
- I set **Microphone Volume** to 80% for clarity
- I set **Speaker Volume** to 60% for comfort
- I press **Mic Test** and confirm my voice sounds clear without distortion

<img width="1008" height="570" alt="Image" src="https://github.com/user-attachments/assets/8fa27772-ad37-48f6-8097-144998e3ec06" />

sheisThanhHien:
<img width="1278" height="718" alt="Image" src="https://github.com/user-attachments/assets/a4bc1971-3b33-4a6d-bed7-8d1ea4877c42" />

huybuidoanquang:
@sheisThanhHien @Luc1404 noise suppression không hoạt động

PR #11910, opened 2026-01-28
260128-web/desktop-voice feature update

Luc1404:
[Desktop/Website] Configure voice settings
[#11868](https://github.com/mezonai/mezon/issues/11868)


libs/components/src/lib/components/SettingVoice/index.tsx
```diff
@@ -0,0 +1,399 @@
...
+
+			recorder.ondataavailable = (e) => {
+				if (e.data && e.data.size > 0) chunksRef.current.push(e.data);
+			};
+
+			recorder.onstop = async () => {
+				const blob = new Blob(chunksRef.current, { type: recorder.mimeType || 'audio/webm' });
+				chunksRef.current = [];
+
+				const url = URL.createObjectURL(blob);
```
huylexuan44: revoke the URL after use or when component unmounts.

huylexuan44: useEffect(() => {
  return () => {
    stopTest();
    if (playbackAudioRef.current?.src) {
      URL.revokeObjectURL(playbackAudioRef.current.src);
    }
  };
}, [stopTest]);

libs/components/src/lib/components/SettingVoice/index.tsx
```diff
@@ -0,0 +1,399 @@
...
+				stopStream();
+			}, 4000);
+		} catch (e) {
+			console.error(e);
+			setIsTesting(false);
+			setTestError(t('setting:voice.errors.micPermission'));
+			cleanupAudioGraph();
+			stopStream();
+		}
+	}, [cleanupAudioGraph, hasSetSinkId, inputDeviceId, micVolume, outputDeviceId, speakerVolume, startLevelMeter, stopStream, stopTest, t]);
```
huylexuan44: using useRef instead of use state without rerender variable

huylexuan44: and break component , keep small logic

libs/components/src/lib/components/SettingVoice/index.tsx
```diff
@@ -0,0 +1,320 @@
...
+import { VolumeSlider } from './VolumeSlider';
+
+const LS_KEYS = {
+	inputDeviceId: 'mezon.voice.inputDeviceId',
+	outputDeviceId: 'mezon.voice.outputDeviceId',
+	micVolume: 'mezon.voice.micVolume',
+	speakerVolume: 'mezon.voice.speakerVolume'
+} as const;
+
+const clamp01 = (n: number) => Math.max(0, Math.min(1, n));
```
huylexuan44, line 15: duplicate code clamp01

Issue #11809, opened 2026-01-22
[Desktop/Website] Topic is not opened when clicking on push notification


sheisThanhHien:
https://github.com/user-attachments/assets/aa60d20d-0ad6-48b8-8d94-a73ef7b54583

sheisThanhHien:
Send sticker ở topic
<img width="1313" height="207" alt="Image" src="https://github.com/user-attachments/assets/efa0c888-1a41-4410-8d36-1782868619fa" />

<img width="381" height="72" alt="Image" src="https://github.com/user-attachments/assets/8cc7ee8c-1b93-41bf-9549-97079885960d" />

PR #11968, opened 2026-01-30
260130-WEB/DESKTOP-fixbug: topic is not opened when clicking push notification

van-datncc:
Task : [Desktop/Website] Topic is not opened when clicking on push notification
[#11809](https://github.com/mezonai/mezon/issues/11809)

### DIFF
--- a/libs/core/src/lib/app/hooks/useMezonNavigateEvent.ts
+++ b/libs/core/src/lib/app/hooks/useMezonNavigateEvent.ts
@@ -1,10 +1,9 @@
-import { getStore, handleTopicNotification, selectDirectMessageEntities, useAppDispatch } from '@mezon/store';
+import { getStore, handleTopicNotification, selectDirectMessageEntities, selectMessageIsLoading, useAppDispatch } from '@mezon/store';
 import { useCallback, useEffect } from 'react';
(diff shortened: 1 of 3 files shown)


libs/core/src/lib/app/hooks/useMezonNavigateEvent.ts
```diff
@@ -23,16 +38,20 @@ export const useMezonNavigateEvent = () => {
...
 					}
 				}
 			}
+
 			navigate(url);
+
 			if (msg) {
-				dispatch(handleTopicNotification({ msg }));
+				setTimeout(() => {
+					dispatch(handleTopicNotification({ msg }));
```
huylexuan44: don't cover all case when channel messages fetch greate than TOPIC_OPEN_DELAY_MS

libs/utils/src/lib/bridge/electron/notification.ts
```diff
@@ -321,17 +322,24 @@ export class MezonNotificationService {
 					existingWindow.focus();
 					const notificationUrl = new URL(link);
```
huylexuan44: add try catch notification URL

huylexuan44 (approved):

Issue #11935, opened 2026-01-29
[Mobile] Lost message in Topic after deleting message


sheisThanhHien:
https://github.com/user-attachments/assets/eee7f091-7242-4bef-8ef4-4243f98d130d

PR #12040, opened 2026-02-03
260203-Mobile-Fix disappear message list when delete IOS

hoangtranlehuy:
Issue: https://github.com/mezonai/mezon/issues/11935
### change visible item for ios

https://github.com/user-attachments/assets/a7d4b512-b550-4a89-acb2-bf5cf998922a

### DIFF
--- a/apps/mobile/src/app/screens/home/homedrawer/components/ChannelMessageList/index.tsx
+++ b/apps/mobile/src/app/screens/home/homedrawer/components/ChannelMessageList/index.tsx
@@ -79,7 +79,7 @@ const ChannelListMessage = React.memo(
 				windowSize={10}
 				onEndReachedThreshold={0.2}
 				maintainVisibleContentPosition={{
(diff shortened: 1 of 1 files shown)


apps/mobile/src/app/screens/home/homedrawer/components/ChannelMessageList/index.tsx
```diff
@@ -79,7 +79,7 @@ const ChannelListMessage = React.memo(
 				windowSize={10}
 				onEndReachedThreshold={0.2}
 				maintainVisibleContentPosition={{
-					minIndexForVisible: Platform.OS === 'ios' ? 2 : 1,
+					minIndexForVisible: Platform.OS === 'ios' ? 0 : 1,
```
phuocnguyenncc: need to add more conditions: check if message length ≤ 2 → return 0

phuocnguyenncc (approved):

PR #12106, opened 2026-02-05
260121 - WEB/DESKTOP-ADMIN - Sort column

huy-ncc-asia:


apps/admin/src/app/components/dashboard/UsersTable.tsx
```diff
@@ -23,12 +26,40 @@ function UsersTable({
...
+	sort,
 	onExportCSV,
 	onToggleColumn,
-	onPageChange
+	onPageChange,
+	onSort
 }: UsersTableProps) {
 	const { t } = useTranslation('dashboard');
 
+	const getSortIcon = (column: string) => {
```
huylexuan44: Code duplication

huy-ncc-asia: I split SortIcon into a separate reusable component.

Issue #12108, opened 2026-02-05
[Community] Clan preview bug in Community


trungdotrongNCC:
Reported by: @hien.nguyenthanh
Time: 2026-02-05T11:20:45.000Z

Description:
trên community, click dô nó ko ra preview clan , nó trắng màn ạ

<img width="1920" height="996" alt="Image" src="https://github.com/user-attachments/assets/2caff35f-55b5-45a9-9b58-44df1cb33363" />

<img width="1920" height="979" alt="Image" src="https://github.com/user-attachments/assets/ade1e021-a42d-4454-841e-cbdf6ea5b610" />

PR #12116, opened 2026-02-06
260206 - WEB- fix26d discover  clan detail dont show 

kind6688:


kind6688:
https://github.com/mezonai/mezon/issues/12108

PR #12172, opened 2026-02-11
260211/WEB/DESKTOP - add Ephemeral chat into voice chat

kind6688:


libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -1132,24 +1134,49 @@ export const sendMessage = createAsyncThunk('messages/sendMessage', async (paylo
...
+				mentions,
+				uploadedFiles,
+				references,
+				anonymous,
+				mentionEveryone,
+				'',
+				code
+			);
+			thunkAPI.dispatch(referencesActions.clearOgpData());
+			return res;
```
huylexuan44: Why do you need return res ? ,

libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -1132,24 +1134,49 @@ export const sendMessage = createAsyncThunk('messages/sendMessage', async (paylo
...
-			code
-		);
-		thunkAPI.dispatch(referencesActions.clearOgpData());
-
-		return res;
+		const currentChannel = selectChannelById(state, channelId);
+		const isVoiceChannel =
+			currentChannel?.type === ChannelType.CHANNEL_TYPE_MEZON_VOICE || currentChannel?.type === ChannelType.CHANNEL_TYPE_STREAMING;
+
+		if (isVoiceChannel) {
```
huylexuan44: move logic to useChatSending, search global sendEphemeralMessage

libs/store/src/lib/messages/messages.slice.ts
kind6688: this is old logic , sir

huylexuan44 (approved):

libs/core/src/lib/chat/hooks/useChatSending.ts
```diff
@@ -117,6 +120,48 @@ export function useChatSending({ mode, channelOrDirect, fromTopic = false }: Use
...
+						mode,
+						isPublic,
+						moreContent,
+						mentions,
+						uploadedFiles,
+						references,
+						false,
+						false,
+						priorityAvatar || '',
+						code
```
huylexuan44, line 159: check code 			TypeMessage.Ephemeral

PR #12178, opened 2026-02-11
260211 - WEB/DESKTOP - fix agent state in voice

sadohpv:
### Changes

- Fix state agent in voice

### DIFF
--- a/libs/components/src/lib/components/VoiceChannel/ControlBar/AgentControl.tsx
+++ b/libs/components/src/lib/components/VoiceChannel/ControlBar/AgentControl.tsx
@@ -1,13 +1,24 @@
+import { useRoomContext } from '@livekit/components-react';
 import { usePermissionChecker } from '@mezon/core';
 import { handleAddAgentToVoice, handleKichAgentFromVoice, selectVoiceInfo, useAppDispatch } from '@mezon/store';
 import { EPermission } from '@mezon/utils';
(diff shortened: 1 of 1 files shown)


libs/components/src/lib/components/VoiceChannel/ControlBar/AgentControl.tsx
```diff
@@ -22,9 +33,36 @@ export const AgentControl = memo(() => {
 		}
 	};
 
-	if (!hasChannelPermission) {
-		return null;
-	}
+	useEffect(() => {
+		if (!room) return;
+
```
huylexuan44, line 37: add check initial 
    if (room.remoteParticipants.get('KOMU')) {
        setOnAgent(true);
    }

libs/components/src/lib/components/VoiceChannel/ControlBar/AgentControl.tsx
huylexuan44: remove setOnAgent

libs/components/src/lib/components/VoiceChannel/ControlBar/AgentControl.tsx
sadohpv: Re check it room when join return empty remoteParticipants

huylexuan44 (approved):

Issue #11558, opened 2026-01-09
[Desktop/Website] Incorrect "member since" date displayed after re-joining clan


sheisThanhHien:
**Steps to reproduce:**
1. User B leaves clan
2. User B re-joins clan 
3. User A opens short profile's user B

<img width="509" height="572" alt="Image" src="https://github.com/user-attachments/assets/5284408c-6a51-4dbf-a790-2d3799894d65" />

<img width="983" height="345" alt="Image" src="https://github.com/user-attachments/assets/a9ba0d38-aa51-4fdf-9360-24517ca0d026" />

PR #12179, opened 2026-02-11
260211-web/desktop-fix bug member since date

Luc1404:
[Desktop/Website] Incorrect "member since" date displayed after re-joining clan
https://github.com/mezonai/mezon/issues/11558
image
image


libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -1090,24 +1089,6 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
 			}
 
 			if (currentClanId === clan_id) {
-				const members = users
```
huylexuan44, line 1093: why do you remove this logic ?

PR #12186, opened 2026-02-11
260211-web/desktop-fix Noise cancellation is not working

Luc1404:
[Desktop/Website] Configure voice settings
https://github.com/mezonai/mezon/issues/11868

### DIFF
--- a/libs/components/src/lib/components/SettingVoice/index.tsx
+++ b/libs/components/src/lib/components/SettingVoice/index.tsx
@@ -3,7 +3,7 @@ import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { DeviceSelector } from './DeviceSelector';
 import { MicTest } from './MicTest';
-import { NoiseSuppressionControl } from './NoiseSuppressionControl';
+import { NoiseSuppressionControl, type NoiseSuppressionControlRef } from './NoiseSuppressionControl';
 import { VolumeSlider } from './VolumeSlider';
 
 const LS_KEYS = {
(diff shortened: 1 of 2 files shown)


libs/components/src/lib/components/SettingVoice/index.tsx
```diff
@@ -27,6 +29,10 @@ interface ISettingVoiceProps {
 export const SettingVoice = ({ menuIsOpen }: ISettingVoiceProps) => {
 	const { t } = useTranslation(['setting']);
 	const [permissionState, setPermissionState] = useState<'unknown' | 'granted' | 'denied'>('unknown');
+	const noiseSuppressionEnabled = useAppSelector(selectNoiseSuppressionEnabled);
```
huylexuan44: don't need selector,  use state and break component for test component

libs/components/src/lib/components/SettingVoice/index.tsx
```diff
@@ -257,14 +357,45 @@ export const SettingVoice = ({ menuIsOpen }: ISettingVoiceProps) => {
...
-	}, [cleanupAudioGraph, hasSetSinkId, startLevelMeter, stopStream, stopTest, t]);
+	}, [cleanupAudioGraph, ensureNoiseSuppressionNode, hasSetSinkId, noiseSuppressionEnabled, startLevelMeter, stopStream, stopTest, t, wireOutput]);
 
 	useEffect(() => {
 		return () => {
 			stopTest();
 		};
 	}, [stopTest]);
 
+	useEffect(() => {
```
huylexuan44: make normal function to handle, don't use hook, 
flow
click event -> call functon enable -> ...

huylexuan44 (approved):

PR #12187, opened 2026-02-11
260211 - WEB/DESKTOP - close ogp message check owner

sadohpv:
### Changes

- Check show delete button when owner message close ogp


apps/chat/src/app/pages/channel/ChannelMedia.tsx
```diff
@@ -9,9 +9,10 @@ import TimelineMessages from './TimelineMessages';
 
 type ChannelMediaProps = {
 	currentChannel: ChannelsEntity | null;
+	accountId?: string;
```
huylexuan44: why do you need use  prop drilling , accountid? you only use selectCurrentUserId in component

huylexuan44 (approved):

Issue #12175, opened 2026-02-11
[Mobile] Lost unread badge when creating private thread and mentioning user


sheisThanhHien:
**Note:** Mention từ mobile ở private thread khi tạo làm mất unread badge ở cả desktop web + mobile

<img width="463" height="449" alt="Image" src="https://github.com/user-attachments/assets/6566c398-6b2b-4afc-8d2d-678c907dbe9b" />

PR #12203, opened 2026-02-12
260212-Mobile-Fix mention to add create private thread

hoangtranlehuy:
Issue: https://github.com/mezonai/mezon/issues/12175
### Change: add members before get send message with mention

https://github.com/user-attachments/assets/21876dcd-70b8-49bb-95ec-6ef914a0c4d3

### DIFF
--- a/libs/core/src/lib/chat/hooks/useThreadMessage.ts
+++ b/libs/core/src/lib/chat/hooks/useThreadMessage.ts
@@ -80,36 +80,32 @@ export function useThreadMessage({ channelId, mode, username }: UseThreadMessage
 				}
 			}
 
-			try {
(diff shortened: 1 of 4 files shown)


libs/core/src/lib/chat/hooks/useThreadMessage.ts
```diff
@@ -80,32 +80,36 @@ export function useThreadMessage({ channelId, mode, username }: UseThreadMessage
...
+				await socket.writeChatMessage(
+					currentClanId,
+					thread.channel_id as string,
+					ChannelStreamMode.STREAM_MODE_THREAD,
+					thread.channel_private === 0,
+					content,
+					mentions,
+					uploadedFiles,
+					references
+				);
```
phuocnguyenncc: mobile send msg http

phuocnguyenncc (approved):

Issue #12199, opened 2026-02-12
[DESKTOP] Cache clan after switching accounts


trungdotrongNCC:
Reported by: @hien.nguyenthanh
Time: 2026-02-12T03:14:37.000Z

Description:
Mezon app login logout 2 account khác nhau bị show clan của user cũ

PR #12230, opened 2026-02-22
260222 - WEB/DESKTOP - try catch call clan desc

sadohpv:
###  Changes

- Add try catch call api clan desc

### Ticket

- [DESKTOP] Cache clan after switching accounts
https://github.com/mezonai/mezon/issues/12199

### DIFF
--- a/libs/store/src/lib/clans/clans.slice.ts
+++ b/libs/store/src/lib/clans/clans.slice.ts
@@ -201,25 +201,31 @@ export const fetchClansCached = async (
 		};
 	}
 
-	const response = await fetchDataWithSocketFallback(
-		ensuredMezon,
-		{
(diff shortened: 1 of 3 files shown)


libs/store/src/lib/clans/clans.slice.ts
```diff
@@ -201,25 +201,31 @@ export const fetchClansCached = async (
...
-		fromCache: false
-	};
+		markApiFirstCalled(apiKey);
+		return {
+			...response,
+			fromCache: false
+		};
+	} catch (error) {
+		return {
+			clandesc: [],
```
huylexuan44: remove

huylexuan44 (approved):

PR #12294, opened 2026-02-26
260226-MOBILE-Update mark as read and badge count logic

thanhlichncc:


libs/store/src/lib/store-mobile.tsx
```diff
@@ -26,6 +26,7 @@ import { notificationReducer } from './notification/notify.slice';
 import { POLICIES_FEATURE_KEY, policiesReducer } from './policies/policies.slice';
 import { reactionReducer } from './reactionMessage/reactionMessage.slice';
 
+import { badgeService } from '@mezon/store-mobile';
```
phuocnguyenncc: check import:

import { badgeService } from './badge/badgeService';

phuocnguyenncc (approved):

Issue #12285, opened 2026-02-25
[Mobile] Lost mention when switching to another channel/thread


sheisThanhHien:
**Steps to reproduce:**
1. At channel A, enter message with mention (role, user, here) into input
2. Back to channel list
3. Go to another channel
4. Back to the channel A
5. Press send message and observe

https://github.com/user-attachments/assets/94dd1786-f19a-4b2b-92a5-d13d60ceb8b0

PR #12301, opened 2026-02-26
260226 - MOBILE - Fix bug mention after back list channel

thanhlichncc:
issue: https://github.com/mezonai/mezon/issues/12285


phuocnguyenncc:
close implement native app

PR #12349, opened 2026-03-03
260303-WEB/DESKTOP-feat: trigger reaction API on disconnect socket

van-datncc:


libs/store/src/lib/reactionMessage/reactionMessage.slice.ts
```diff
@@ -161,32 +161,59 @@ export const writeMessageReaction = createAsyncThunk(
...
-								topic_id || '0',
-								emoji_recent_id || '0',
-								sender_name
-							),
-							2000,
-							'Message reaction operation timed out'
+				let socketSuccess = false;
+				if (socket) {
+					try {
+						await retryWithBackoff(
```
huylexuan44, line 167: use socketState.isConnected to check

huylexuan44 (approved):

PR #12475, opened 2026-03-12
260312-WEB/DESKTOP - desktop-add-friend

kind6688:
https://github.com/mezonai/mezon/issues/12466


libs/store/src/lib/friends/friend.slice.ts
```diff
@@ -221,26 +221,20 @@ export const sendRequestUnblockFriend = createAsyncThunk('friends/requestUnblock
 	return response;
 });
 
-export const upsertFriendRequest = createAsyncThunk(
```
huylexuan44, line 224: - Sender spam API not yet processed 
- disable button add friend when api call

Issue #12283, opened 2026-02-25
[Desktop/Website] Add btn create poll


sheisThanhHien:

sheisThanhHien:
## 1/ Theme Light, Sunrise, 
- Sửa dấm chấm trắng selected thành màu xanh (1)
- Text sau khi vote bị chìm, ko được để màu xám (2)

<img width="663" height="415" alt="Image" src="https://github.com/user-attachments/assets/f12c68ee-2d42-40b3-89bb-5d335e79a5d9" />

<img width="603" height="427" alt="Image" src="https://github.com/user-attachments/assets/34b24810-d83f-4e6c-b281-d24ef43356b9" />


## 2/ Tạo poll thêm emoji không thấy hiển thị emoji (trao đổi với BE)

<img width="480" height="508" alt="Image" src="https://github.com/user-attachments/assets/ffa9fbe3-1601-4b40-a9d7-6573fe8b01b6" />


## 3/ 1 days mà có `s`?

<img width="614" height="328" alt="Image" src="https://github.com/user-attachments/assets/9cc06f49-de4b-4f07-a9a9-c7c35af6def2" />

## 4/ 1 option thì phải disable nút Post (ít nhất 2 options mới đc tạo poll) 

<img width="598" height="565" alt="Image" src="https://github.com/user-attachments/assets/fca79b8c-0cb1-44cb-93fb-035c3bb6bd7c" />

## 5/ Đôi lúc tạo poll bị 400

<img width="1766" height="930" alt="Image" src="https://github.com/user-attachments/assets/17fc18a4-3cb4-4744-9caa-ed46eba70c43" />

## 6/ Ghim cái poll

<img width="859" height="611" alt="Image" src="https://github.com/user-attachments/assets/8749712d-5e14-48a7-a7c2-6c9a3504eedc" />

## 7/ Ẩn những options này ở context menu đối với poll

<img width="1032" height="613" alt="Image" src="https://github.com/user-attachments/assets/cf474d2d-40a5-43c2-8f3a-958d767534ae" />


## 8/ Không tạo được poll ở DM group chat

sheisThanhHien:
### 1/ Chặn Edit, Create Thread, Create Topic, Forward poll

<img width="606" height="134" alt="Image" src="https://github.com/user-attachments/assets/39b85d15-0828-4231-8885-7e7c17922c89" />

[...]

PR #12521, opened 2026-03-16
260316 - WEB/DESKTOP - update logic poll message

Luc1404:


Luc1404:
[Desktop/Website] Add btn create poll
https://github.com/mezonai/mezon/issues/12283

libs/components/src/lib/components/MessageWithUser/PollMessage.tsx
```diff
@@ -115,13 +216,24 @@ export const PollMessage = ({
 				const voters: PollVoter[] = [];
 
 				userIds.forEach((userId) => {
+					const state = getStore().getState();
```
huylexuan44, line 219: move to outside foreach

libs/store/src/lib/polls/polls.slice.ts
```diff
@@ -113,10 +106,30 @@ export const getPoll = createAsyncThunk<ApiGetPollResponse, ApiGetPollRequest>('
 	}
 });
```
huylexuan44, line 108: remove pool slice

PR #12542, opened 2026-03-17
260317 - WEB/DESKTOP - fix fetch message with id

sadohpv:
### Changes

- Fetch message with id

### DIFF
--- a/libs/store/src/lib/messages/messages.slice.ts
+++ b/libs/store/src/lib/messages/messages.slice.ts
@@ -222,7 +222,7 @@ export const fetchMessagesCached = async (
 		}
 	}
 	const channelData = state[MESSAGES_FEATURE_KEY].channelMessages[channelId];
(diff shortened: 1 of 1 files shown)


libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -375,13 +375,14 @@ export const fetchMessages = createAsyncThunk(
 			if (!currentUser) {
 				currentUser = await thunkAPI.dispatch(accountActions.getUserProfile()).unwrap();
 			}
+			const lastMessageId = selectLastMessageIdByChannelId(state, channelId);
```
huylexuan44, line 378: If a message is deleted, will it have any effect?

sadohpv: I select last messageId in store by selectMessageIdsByChannelId.at(-1)

huylexuan44 (approved):

Issue #12546, opened 2026-03-17
[Desktop/Website] UI bug embed in Search


sheisThanhHien:
<img width="709" height="678" alt="Image" src="https://github.com/user-attachments/assets/0d590e0d-7b66-49be-b7cc-fce8f77273c9" />

PR #12549, opened 2026-03-17
260317-web/desktop-fix UI bug embed

Luc1404:
[Desktop/Website] UI bug embed in Search
https://github.com/mezonai/mezon/issues/12546


libs/components/src/lib/components/EmbedMessage/components/EmbedImage.tsx
```diff
@@ -7,20 +7,14 @@ interface EmbedImageProps {
 	height?: number;
 }
 
-export function EmbedImage({ url, width, height }: EmbedImageProps) {
+export function EmbedImage({ url }: EmbedImageProps) {
```
huylexuan44: height and width cannot be deleted.

huylexuan44: try use aspect-ratio

huylexuan44 (approved):

PR #12577, opened 2026-03-20
260320 - WEB/DESKTOP/MOBILE - add español language

Luc1404:


libs/components/src/lib/components/SettingLanguage/index.tsx
```diff
@@ -28,6 +28,11 @@ export const SettingLanguage = ({ menuIsOpen }: ISettingLanguageProps) => {
 				title: `${t('setting:language.russian')}`,
 				value: 'ru',
 				contributed: 'tsruber'
+			},
```
huylexuan44, line 31: if (id.includes('libs/translations/src/languages/en')) {
							return 'i18n-en';
						}

add to vite config chunk mezon/apps/chat/vite.config.ts

libs/translations/src/languages/en/setting.json
```diff
@@ -85,7 +85,8 @@
 		"description": "Select your preferred language.",
 		"english": "English",
```
huylexuan44, line 86: es folder missing file setting.json

PR #12578, opened 2026-03-20
260320 - WEB/DESKTOP - ref clan badge

sadohpv:
### Changes

- Changes clan store badge

### DIFF
--- a/libs/store/src/lib/clans/clans.slice.ts
+++ b/libs/store/src/lib/clans/clans.slice.ts
@@ -48,8 +48,9 @@ interface ClanMeta {
 }
 
 interface ClanUnreadState {
-	id: string; // clanId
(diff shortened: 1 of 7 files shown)


libs/store/src/lib/clans/clans.slice.ts
huylexuan44: Badge data was moved out of state.entities into state.clanUnreadStates

libs/store/src/lib/clans/clans.slice.ts
```diff
@@ -48,8 +48,9 @@ interface ClanMeta {
 }
 
 interface ClanUnreadState {
```
huylexuan44, line 50: setClanBadgeCount change to upsertOne

huylexuan44 (approved):

PR #12596, opened 2026-03-21
260320 - MOBILE - bug26a/mobile/update direct message bagge order

hoangtranlehuy:
bug26a/mobile/update direct message bagge order
### Change check mobile when get list badge DM.

### DIFF
--- a/libs/store/src/lib/clans/clans.slice.ts
+++ b/libs/store/src/lib/clans/clans.slice.ts
@@ -501,7 +501,7 @@ interface JoinClanPayload {
 	isMobile?: boolean;
 }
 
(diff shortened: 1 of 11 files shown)


libs/store/src/lib/clans/clans.slice.ts
```diff
@@ -493,15 +498,16 @@ export const updateUser = createAsyncThunk(
 
 interface JoinClanPayload {
 	clanId: string;
+	isMobile?: boolean;
 }
 
-export const joinClan = createAsyncThunk<void, JoinClanPayload>('direct/joinClan', async ({ clanId }, thunkAPI) => {
+export const joinClan = createAsyncThunk<void, JoinClanPayload>('direct/joinClan', async ({ clanId, isMobile }, thunkAPI) => {
```
phuocnguyenncc: isMobile = false

Issue #12491, opened 2026-03-13
[Desktop/Website] Edited category is not updated in System message channels setting


sheisThanhHien:
**Steps to reproduce:**
1. Edit category A to A1
2. Go to clan settings
3. Observe the category which is updated to A1

**Actual Result:** Stills display old non-edited category name

<img width="1085" height="795" alt="Image" src="https://github.com/user-attachments/assets/ef9a15d3-30e1-4d12-af75-1b52b6cfb195" />

PR #12609, opened 2026-03-23
260323 - WEB/DESKTOP - Edited category is not updated in System message channels setting

Luc1404:
[Desktop/Website] Edited category is not updated in System message channels setting
https://github.com/mezonai/mezon/issues/12491

### DIFF
--- a/libs/store/src/lib/channels/channels.slice.ts
+++ b/libs/store/src/lib/channels/channels.slice.ts
@@ -1046,6 +1046,19 @@ export const channelsSlice = createSlice({
 			}
 		},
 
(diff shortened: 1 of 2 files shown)


libs/store/src/lib/channels/channels.slice.ts
```diff
@@ -1046,6 +1046,19 @@ export const channelsSlice = createSlice({
 			}
 		},
```
huylexuan44, line 1048: Missing: channelsActions.updateCategoryNameForChannels

huylexuan44 (approved):

PR #12612, opened 2026-03-23
230326-Mobile-Fix badge direct message and last sent content

hoangtranlehuy:
Issues: https://github.com/mezonai/mezon/issues/12576
### Change:
- Update last sent timestamp channel meta
- Update direct meta when get badge clan to show last send messages content
- Write badge count to direct adapter.


apps/mobile/src/app/screens/home/homedrawer/components/UnreadDMBadgeList/index.tsx
```diff
@@ -24,6 +33,10 @@ const UnreadDMBadgeItem = memo(({ dmId, numUnread }: { dmId: string; numUnread:
 	const isTabletLandscape = useTabletLandscape();
 	const dispatch = useAppDispatch();
 
+	if (isEmpty(dm)) {
+		return null;
+	}
```
phuocnguyenncc: Crash: Rules of Hooks
React Hooks must be called in the same order in every component render.

phuocnguyenncc (approved):

PR #12747, opened 2026-04-03
260403 - WEB/DESKTOP - Fix OGP message size limit

sadohpv:
### Changes

- Fix ogp size and UX


libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -1165,6 +1167,8 @@ export const sendMessage = createAsyncThunk('messages/sendMessage', async (paylo
 				throw new Error('Socket not connected');
 			}
 		} catch (err) {
```
huylexuan44, line 1167: remove log

huylexuan44, line 1167: remove log

huylexuan44 (approved):

Issue #12725, opened 2026-04-02
[Desktop/Website] Error displaying user in member list after adding


sheisThanhHien:
### Env: DEV
add user dô group chat bị như hình
<img width="310" height="740" alt="Image" src="https://github.com/user-attachments/assets/44d8f829-407b-4813-8626-945503b5d44e" />

PR #12762, opened 2026-04-06
260406 - WEB/DESKTOP - Error displaying user in member list after adding

Luc1404:
[Desktop/Website] Error displaying user in member list after adding
https://github.com/mezonai/mezon/issues/12725


libs/store/src/lib/permissionChannel/permissionChannel.slice.ts
```diff
@@ -26,6 +26,13 @@ export const addChannelUsers = createAsyncThunk(
 			}
 			if (channelId && channelType) {
 				thunkAPI.dispatch(userChannelsActions.addUserChannel({ channelId, userAdds: userIds }));
+				await thunkAPI.dispatch(
+					userChannelsActions.fetchUserChannels({
+						channelId,
+						noCache: true,
```
huylexuan44: update addUserChannel instead of call api

PR #12935, opened 2026-04-29
260429 - WEB/DESKTOP - create only option archived channel

Luc1404:


libs/components/src/lib/components/PanelChannel/index.tsx
```diff
@@ -216,6 +217,54 @@ const PanelChannel = ({ coords, channel, openSetting, setIsShowPanelChannel, onD
...
+			);
+			if (channel.id === currentChannelId || channel.id === selectedChannel) {
+				if (isThread) {
+					navigate(`/chat/clans/${currentClanId}/channels/${channel.parent_id}`);
+				} else {
+					navigate(`/chat/clans/${currentClanId}/channels/${welcomeChannelId}`);
+				}
+			}
+		} catch (error) {
+			// Ignored intentionally or handled globally
```
huylexuan44, line 238: add toast fail

libs/store/src/lib/channels/channels.slice.ts
```diff
@@ -1508,6 +1522,18 @@ export const channelsSlice = createSlice({
 				state.error = action.error.message;
 			});
 
+		builder
+			.addCase(archiveChannel.pending, (state: ChannelsState) => {
```
huylexuan44, line 1526: remove all

Issue #13022, opened 2026-05-11
[Desktop/Website] Add member count under the header of group name


sheisThanhHien:
<img width="137" height="58" alt="Image" src="https://github.com/user-attachments/assets/473ddc61-6def-41a0-9f31-c09f34266271" />

sheisThanhHien:
<img width="550" height="275" alt="Image" src="https://github.com/user-attachments/assets/8047bbd3-9fbe-482c-9d69-9277014c7d35" />

PR #13025, opened 2026-05-11
260511 - WEB/DESKTOP - fix status and group member count

sadohpv:
### Changes

- Fix status custom and group member count


sheisThanhHien:
fix #13022

PR #13081, opened 2026-05-14
260514 - WEB/DESKTOP - fix event panel

sadohpv:
### Changes

- Changes condition show option event

### DIFF
--- a/libs/store/src/lib/store.tsx
+++ b/libs/store/src/lib/store.tsx
@@ -443,17 +443,7 @@ const thunkNameLogger = () => (next: any) => (action: any) => {
 	if (isThunk) {
 		const [slice, actionName, status] = action.type.split('/');
 		if (status === 'pending') {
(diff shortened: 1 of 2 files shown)


libs/store/src/lib/store.tsx
```diff
@@ -443,17 +443,7 @@ const thunkNameLogger = () => (next: any) => (action: any) => {
...
-			});
-		}
-
-		if (status === 'fulfilled') {
-			console.warn(`✅ THUNK SUCCESS: ${slice}/${actionName}`);
-		}
-
-		if (status === 'rejected') {
-			console.warn(`❌ THUNK ERROR: ${slice}/${actionName}`, action.error);
+			console.warn(`🚀 : ${slice}/${actionName}`);
```
huylexuan44, line 446: remove log

sadohpv: Log on dev only

huybuidoanquang (approved):

Issue #12705, opened 2026-03-31
[Desktop/Website] Display option `Unblock` of user has been blocked


sheisThanhHien:
**Env:** prod, dev

**Steps to reproduce:**
1. User A blocks user B
2. User B right-click on DM's user A --> **_Observe the menu_**
3. Refresh app, repeat step 2 --> Observe the menu

**Actual Result:** Option `Unblock` is showed in menu of user B who has been blocked, f5 is OKE

https://github.com/user-attachments/assets/c61e7777-6f07-430f-abab-67b03c96e546

PR #13100, opened 2026-05-19
260519 - WEB/DESKTOP -  Display option Unblock of user has been blocked

Luc1404:
[Desktop/Website] Display option Unblock of user has been blocked
https://github.com/mezonai/mezon/issues/12705

### DIFF
--- a/libs/core/src/lib/chat/contexts/ChatContext.tsx
+++ b/libs/core/src/lib/chat/contexts/ChatContext.tsx
@@ -2445,7 +2445,6 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
 			const targetUserId = blockFriend.user_id;
 
 			if (targetUserId === myId) {
-				void dispatch(friendsActions.fetchListFriends({ noCache: true }));
 				return;
 			}
 
--- a/libs/core/src/lib/chat/contexts/ChatContext.tsx
+++ b/libs/core/src/lib/chat/contexts/ChatContext.tsx
@@ -74,6 +74,7 @@ import {
 	selectDirectById,
 	selectDmGroupCurrentId,
 	selectDmMetaEntities,
+	selectFriendById,
 	selectIsInCall,
 	selectLastMessageByChannelId,
 	selectLastSentMessageStateByChannelId,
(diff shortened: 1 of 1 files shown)


libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -2437,16 +2438,22 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
...
-			);
+			const myId = userId as string;
+			const targetUserId = blockFriend.user_id;
+
+			if (targetUserId === myId) {
+				void dispatch(friendsActions.fetchListFriends({ noCache: true }));
+				return;
+			}
+
+			const existing = selectFriendById(getStore().getState() as RootState, targetUserId);
```
huylexuan44, line 2451: getStore snapshot may race local applyFriendBlockState — can misread iBlockedThem and refetch or skip wrongly.

libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -2437,16 +2438,22 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
...
+			const iBlockedThem = existing?.state === EStateFriend.BLOCK && existing.source_id === myId;
+			if (!iBlockedThem) {
```
huylexuan44, line 2453: When iBlockedThem is true the handler no-ops — confirm socket never arrives before optimistic block state is in the store.

libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -2437,16 +2438,22 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
...
+				void dispatch(friendsActions.fetchListFriends({ noCache: true }));
```
huylexuan44, line 2454: fetchListFriends noCache on every remote block — OK for correctness; debounce if block events can burst.

libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -2437,16 +2438,22 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
...
-				friendsActions.applyFriendBlockState({
-					userId: blockFriend.user_id,
-					state: EStateFriend.BLOCK,
-					sourceId: userId as string
-				})
-			);
+			const myId = userId as string;
+			const targetUserId = blockFriend.user_id;
+
+			if (targetUserId === myId) {
```
huylexuan44, line 2447: targetUserId === myId refetches only — confirm socket payload uses this shape when you are the blocked user.

huylexuan44 (approved):

PR #13129, opened 2026-05-20
260520 - HOTFIX - probe video codec incrementally for MOV on Electron Mac

huylexuan44:
Use ranged MP4Box probing with smaller initial fetch, QuickTime detection, and clearer unsupported codec messaging on macOS desktop.

### DIFF
--- a/libs/components/src/lib/components/MessageWithUser/MessageVideo.tsx
+++ b/libs/components/src/lib/components/MessageWithUser/MessageVideo.tsx
@@ -559,7 +559,7 @@ function DefaultVideo({ attachmentData, isMobile = false, isPreview = false, obs
 }
 
 function MessageVideo(props: MessageImage) {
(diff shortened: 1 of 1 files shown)


huylexuan44:
H.264 .mov — plays inline after probe (no false “unsupported”).
HEVC / Dolby Vision .mov — blocked with download CTA + codec line when probe succeeds.
.mp4 (iPhone HEVC, no .mov ext) — still blocked if probe parses codec; not silently broken.
Scroll away mid-probe — no spinner stuck; no extra network after leaving viewport.
Slow / flaky network — timeout (~5s) → MOV shows error, not infinite spinner.
CDN without Range (200 not 206) — .mov must not false-block if <video> can play; or confirm intentional block.
Large MOV (moov near end, within 512 KB) — probe completes and correct play/block decision.
Non-Mac / web chat — unchanged; videos still play normally.

PR #13135, opened 2026-05-21
260521 - WEB/DESKTOP - fix marks as read channel

sadohpv:
### Changes

- Change filter to get channel not just only thread

### Description
- Old logic get count from channelEntity. New logic haft years ago get unread and count_unread from channelMeta 
- Use channelMeta instead of channelEntity to get badge count for case markAsRead

### Test Case

- Mark as read Clan 
- Mark as read Channel 
- Mark as read Cate 
- Refresh for each case mark as read

### DIFF
--- a/libs/core/src/lib/chat/contexts/ChatContext.tsx
+++ b/libs/core/src/lib/chat/contexts/ChatContext.tsx
@@ -2509,7 +2509,7 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
 		badgeService.markAsReadChannel(
 			markAsReadEvent.clan_id as string,
 			markAsReadEvent.channel_id,
-			[markAsReadEvent.channel_id, ...channelIds],
+			[...channelIds],
 			channelUpdates,
 			relatedChannels.map((channel) => ({
 				channelId: channel.id,
--- a/libs/core/src/lib/chat/contexts/ChatContext.tsx
+++ b/libs/core/src/lib/chat/contexts/ChatContext.tsx
@@ -2496,12 +2496,16 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
 			badgeService.markAsReadCategory(markAsReadEvent.clan_id as string, markAsReadEvent.category_id, channelIds, channelUpdates);
 			return;
 		}
-		const relatedChannels = channels.filter((channel) => channel.parent_id === markAsReadEvent.channel_id);
+		const relatedChannels = channels.filter(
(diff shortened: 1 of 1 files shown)


libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -2496,20 +2496,24 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
...
 		const channelIds = relatedChannels.map((channel) => channel.id);
 		const channelUpdates = channelIds.map((channelId) => ({
 			channelId,
 			messageId: selectLatestMessageId(store.getState(), channelId) || undefined
 		}));
+
 		badgeService.markAsReadChannel(
 			markAsReadEvent.clan_id as string,
 			markAsReadEvent.channel_id,
 			[markAsReadEvent.channel_id, ...channelIds],
```
huylexuan44: Parent is now in relatedChannels but channelIds also prepends markAsReadEvent.channel_id — dedupe to avoid duplicate reset entries.

libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -2496,20 +2496,24 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
...
 			channelUpdates,
 			relatedChannels.map((channel) => ({
 				channelId: channel.id,
-				count: (channel.count_mess_unread ?? 0) * -1
+				count: (channelMeta?.[channel.id]?.count_mess_unread ?? 0) * -1
```
huylexuan44, line 2516: channelMeta count aligns with badge store; confirm clan/category mark-as-read paths do not need the same source.

libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -2496,20 +2496,24 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
 			badgeService.markAsReadCategory(markAsReadEvent.clan_id as string, markAsReadEvent.category_id, channelIds, channelUpdates);
 			return;
 		}
-		const relatedChannels = channels.filter((channel) => channel.parent_id === markAsReadEvent.channel_id);
+		const relatedChannels = channels.filter(
+			(channel) => channel.parent_id === markAsReadEvent.channel_id || channel.channel_id === markAsReadEvent.channel_id
```
huylexuan44, line 2500: Filter now includes parent channel — threadIds flatMap below may also run on parent.threadIds; verify lastSeen is not duplicated.

huylexuan44 (approved):

PR #13175, opened 2026-05-27
260527 - WEB/DESKTOP - fix clan name in search

sadohpv:
### Changes

- Fix clan name on search 

### Description 
- BE has cache clan_name on data of allChannelOfAllClan api for search in ctrl K 
- Dont use data clan_name on api. Use on clanListDesc API call before
- Same issue in share contact modal

### Test 

- Change clan name then ctrl K 
- Ctrl R then ctrl K
- Share contact modal
-


libs/components/src/lib/components/SearchModal/index.tsx
```diff
@@ -93,7 +95,7 @@ function SearchModal({ onClose }: SearchModalProps) {
 				list.push({
 					id: item?.channel_id ?? '',
 					name: item?.channel_label ?? '',
-					subText: item?.clan_name ?? '',
+					subText: item?.clan_id ? listClans?.[item.clan_id]?.clan_name || '' : item.clan_name || '',
```
huylexuan44, line 147: Regression risk: if clan entity is missing, this clears label; fallback to item.clan_name.

libs/components/src/lib/components/ShareContact/useShareContact.ts
```diff
@@ -117,12 +119,12 @@ export const useShareContact = ({ contactUser, t }: UseShareContactProps) => {
 				type: (ch.type === ChannelType.CHANNEL_TYPE_THREAD ? 'thread' : 'channel') as ShareItemType,
 				channelId: ch.channel_id || ch.id,
 				clanId: ch.clan_id || '',
-				clanName: ch.clan_name || '',
+				clanName: ch.clan_id ? clansEntities?.[ch.clan_id]?.clan_name || '' : '',
```
huylexuan44, line 123: Regression risk: missing clan entity now blanks name; keep fallback to ch.clan_name.

libs/components/src/lib/components/SearchModal/index.tsx
sadohpv: Can not fallback to item.clan_name because it wrong data and BE will not response it in future

libs/components/src/lib/components/ShareContact/useShareContact.ts
sadohpv: Can not fallback to ch.clan_name because it wrong data and BE will not response it in future

huylexuan44 (approved):

PR #13235, opened 2026-06-03
260603 - WEB/DESKTOP - fix get lastMessage from badgeCount

sadohpv:
### Changes

- get last message of channel from listChannelBadgeCount 

### Description 

- last message will not response from listChannelMessages anymore so get from another api

### DIFF
--- a/libs/store/src/lib/clans/clans.slice.ts
+++ b/libs/store/src/lib/clans/clans.slice.ts
@@ -178,13 +178,10 @@ export const listChannelBadgeCount = createAsyncThunk('clans/listChannelBadgeCou
 			thunkAPI.dispatch(channelMetaActions.updateBulkChannelMetadata({ data: response?.channeldesc as ChannelMetaEntity[], clanId }));
 
 			const listLastMessage = (response.channeldesc ?? []).reduce<ApiChannelMessageHeaderWithChannel[]>((acc, channel) => {
-				if (channel.channel_id) {
+				if (channel.channel_id && channel.last_seen_message) {
 					acc.push({
-						channel_id: channel.channel_id,
-						content: channel.last_sent_message?.content,
-						id: channel.last_sent_message?.id,
-						sender_id: channel.last_sent_message?.sender_id,
-						timestamp_seconds: channel.last_seen_message?.timestamp_seconds
+						...channel.last_sent_message,
+						channel_id: channel.channel_id
 					});
 				}
 				return acc;
--- a/libs/store/src/lib/clans/clans.slice.ts
+++ b/libs/store/src/lib/clans/clans.slice.ts
@@ -1,5 +1,5 @@
 import { captureSentryError } from '@mezon/logger';
-import type { IClan, LoadingStatus } from '@mezon/utils';
+import type { ApiChannelMessageHeaderWithChannel, IClan, LoadingStatus } from '@mezon/utils';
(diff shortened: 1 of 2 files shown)


libs/store/src/lib/clans/clans.slice.ts
```diff
@@ -173,8 +174,23 @@ export const listChannelBadgeCount = createAsyncThunk('clans/listChannelBadgeCou
...
+			thunkAPI.dispatch(channelMetaActions.updateBulkChannelMetadata({ data: response?.channeldesc as ChannelMetaEntity[], clanId }));
+
+			const listLastMessage = (response.channeldesc ?? []).reduce<ApiChannelMessageHeaderWithChannel[]>((acc, channel) => {
+				if (channel.channel_id) {
+					acc.push({
+						channel_id: channel.channel_id,
+						content: channel.last_sent_message?.content,
+						id: channel.last_sent_message?.id,
+						sender_id: channel.last_sent_message?.sender_id,
+						timestamp_seconds: channel.last_seen_message?.timestamp_seconds
```
huylexuan44: timestamp_seconds uses last_seen_message but id/content/sender_id use last_sent_message — use last_sent_message timestamp.

huylexuan44:
## Review summary

- **High:** `timestamp_seconds` is taken from `last_seen_message` while other fields come from `last_sent_message` — likely wrong preview time / sort order.
- **Medium:** Entries are pushed for any `channel_id` even when `last_sent_message` is missing; may store partial last-message headers.
- **Note:** `setManyLastMessages` runs only when `!checkJoinList[clanId]` (same gate as metadata) — OK for join/DM bootstrap; confirm other paths still seed previews for existing sessions.

Verdict: Request changes — fix timestamp source and guard empty last_sent_message before merge.

libs/store/src/lib/clans/clans.slice.ts
```diff
@@ -173,8 +174,23 @@ export const listChannelBadgeCount = createAsyncThunk('clans/listChannelBadgeCou
 			'channel_badge_count'
 		);
 
-		if ((response as any)?.channeldesc && clanId && !state.clans.checkJoinList[clanId]) {
-			thunkAPI.dispatch(channelMetaActions.updateBulkChannelMetadata({ data: (response as any)?.channeldesc, clanId }));
+		if (response.channeldesc && clanId && !state.clans.checkJoinList[clanId]) {
+			thunkAPI.dispatch(channelMetaActions.updateBulkChannelMetadata({ data: response?.channeldesc as ChannelMetaEntity[], clanId }));
+
+			const listLastMessage = (response.channeldesc ?? []).reduce<ApiChannelMessageHeaderWithChannel[]>((acc, channel) => {
+				if (channel.channel_id) {
```
huylexuan44: Push only when last_sent_message exists; empty headers can overwrite lastMessageByChannel for a channel_id.

libs/store/src/lib/clans/clans.slice.ts
```diff
@@ -173,8 +174,23 @@ export const listChannelBadgeCount = createAsyncThunk('clans/listChannelBadgeCou
...
+					acc.push({
```
huylexuan44, line 182: Match channels.slice: spread channel.last_sent_message and add channel_id instead of hand-picking fields.

huylexuan44 (approved):

Issue #13287, opened 2026-06-12
[Desktop/Website] Send message in weak connection sorts incorrectly


sheisThanhHien:
<img width="413" height="294" alt="Image" src="https://github.com/user-attachments/assets/cfbed0ff-e74c-46c0-b6fa-7374b16ff04c" />

khi gui anh va kem text thi bi loi. Sau do message loi se thay doi theo message cuoi cung cua minh
ko chuot phai vao duoc luon
luc day ko update duoc lsent
mạng kém mới bị

PR #13289, opened 2026-06-12
260612 - WEB/DESKTOP - remove fake message after fail

sadohpv:
### Changes

- Remove message when end error catch 

### Ticket 

- [Desktop/Website] Send message in weak connection sorts incorrectly
https://github.com/mezonai/mezon/issues/13287

### Test 
- Send fail message when off connection 
- Send normal 
- Send fail then connect success send normal

### DIFF
--- a/libs/store/src/lib/messages/messages.slice.ts
+++ b/libs/store/src/lib/messages/messages.slice.ts
@@ -1401,14 +1401,13 @@ export const sendMessage = createAsyncThunk('messages/sendMessage', async (paylo
 				);
 			}
 		} catch (error) {
-			if (error instanceof Error && error.name === 'SocketTimeoutError') {
+			if (error instanceof Error && (error.name === 'SocketTimeoutError' || error.name === 'SendTimeoutError')) {
 				thunkAPI.dispatch(
 					messagesActions.remove({
 						messageId: fakeMessage.id,
 						channelId: fakeMessage.channel_id
 					})
 				);
-				return;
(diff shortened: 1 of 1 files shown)


huylexuan44 (approved):

libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -1402,6 +1402,12 @@ export const sendMessage = createAsyncThunk('messages/sendMessage', async (paylo
 			}
 		} catch (error) {
 			if (error instanceof Error && error.name === 'SocketTimeoutError') {
```
huylexuan44: Only SocketTimeoutError removes the fake; SendTimeoutError (30s race) still markAsError — weak network may hit either path.

libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -1402,6 +1402,12 @@ export const sendMessage = createAsyncThunk('messages/sendMessage', async (paylo
...
+				thunkAPI.dispatch(
+					messagesActions.remove({
+						messageId: fakeMessage.id,
+						channelId: fakeMessage.channel_id
+					})
+				);
 				return;
```
huylexuan44: Early return skips sendTimeoutMap/timeoutId cleanup; clear the pending timer here like the branch below.

huylexuan44 (changes requested):

Issue #13281, opened 2026-06-11
[Desktop/Website] Stream channel disappears after removing thumbnail


sheisThanhHien:
## Bug 1: i18n

<img width="1347" height="776" alt="Image" src="https://github.com/user-attachments/assets/45515d6a-4a46-41a4-a131-93e6a98e81f0" />

## Bug 2: Stream channel disappears after removing thumbnail
**Steps to reproduce:**
1. Go to Edit stream channel
2. Click Stream Thumbnail
3. Upload a thumbnail
4. After uploading thumbnail, click remove thumbnail 

**Actual Result:**
- The user is kicked back out of the thumbnail editing screen
- The stream channel disappears in side bar channel list

**Evidence:**

https://github.com/user-attachments/assets/ada2f351-a0d8-48c3-81d0-28f3e2190ef1

sheisThanhHien:
Remove stream thumbnail lỗi
Log: {"level":"error","ts":"2026-06-29T15:09:19.101+0700","caller":"server/core_channel.go:752","msg":"Invalid image URL","url":"0","error":"validation failed for field 'image_url'"}

PR #13296, opened 2026-06-15
260615 - WEB/DESKTOP - Stream channel disappears after removing thumbnail

Luc1404:
[Desktop/Website] Stream channel disappears after removing thumbnail
https://github.com/mezonai/mezon/issues/13281

### DIFF
--- a/libs/components/src/lib/components/ChannelSetting/Component/StreamThumbnail/index.tsx
+++ b/libs/components/src/lib/components/ChannelSetting/Component/StreamThumbnail/index.tsx
@@ -2,6 +2,7 @@
 import {
 	channelsActions,
 	selectChannelById,
+	selectCurrentChannelCategoryId,
 	selectCurrentChannelClanId,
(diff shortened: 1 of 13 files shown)


libs/components/src/lib/components/ChannelSetting/Component/StreamThumbnail/index.tsx
```diff
@@ -162,7 +162,7 @@ const StreamThumbnailChannel = (props: StreamThumbnailChannelProps) => {
 				channelsActions.updateChannel({
 					channel_id: channelId,
 					channel_label: channelLabel,
-					category_id: undefined,
+					category_id: currentChannel?.category_id,
```
huylexuan44: Use category_id fallback chain (currentChannel ?? channel) like clanId/channelLabel — avoid undefined if store entry is missing.

huylexuan44 (approved):

PR #13298, opened 2026-06-15
260615 - WEB/DESKTOP - fix connect socket admin

sadohpv:
### Changes 
- remove code not use in develop

### DIFF
--- a/apps/admin/src/app/loader/applicationLoader.ts
+++ b/apps/admin/src/app/loader/applicationLoader.ts
@@ -1,4 +1,4 @@
-import { fetchMezonOauthClient, getApplicationDetail, getStoreAsync, selectApplicationById, setCurrentAppId } from '@mezon/store';
+import { fetchMezonOauthClient, getApplicationDetail, getStoreAdminAsync, selectApplicationById, setCurrentAppId } from '@mezon/store';
 import type { CustomLoaderFunction } from './appLoader';
 
 interface IBotLoaderData {
@@ -12,7 +12,7 @@ export const applicationLoader: CustomLoaderFunction = async ({ params, dispatch
 		throw new Error('Application ID null');
 	}
 
-	const store = await getStoreAsync();
+	const store = await getStoreAdminAsync();
 
 	const currentApp = selectApplicationById(store.getState(), applicationId);
 	if (!currentApp) {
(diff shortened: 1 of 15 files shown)


apps/admin/src/app/loader/applicationLoader.ts
```diff
@@ -21,11 +13,6 @@ export const applicationLoader: CustomLoaderFunction = async ({ params, dispatch
 	}
 
 	const store = await getStoreAsync();
```
huylexuan44: getStoreAsync waits for initStore in store.tsx; admin uses initAdminStore — loaders may timeout or read the wrong store.

apps/admin/src/app/loader/authLoader.ts
```diff
@@ -24,11 +23,7 @@ function getRedirectTo(initialPath?: string): string {
 
 export const authLoader: CustomLoaderFunction = async ({ dispatch, initialPath }) => {
 	try {
-		const profileResponse = await dispatch(accountActions.getUserProfile());
 		await dispatch(fetchApplications({}));
```
huylexuan44, line 26: getUserProfile was removed but nothing sets auth.isLogin; RootLayout still gates on selectIsLogin.

libs/store/src/lib/auth/auth.slice.ts
```diff
@@ -399,6 +399,11 @@ export const authSlice = createSlice({
 					state.session = value;
 				}
 			}
+		},
+		setSessionToken(state, action: PayloadAction<{ token?: string; refresh_token?: string }>) {
+			if (state.session) {
```
huylexuan44, line 407: setSessionToken no-ops when state.session is null — SessionRefresh tokens are dropped after a successful refresh.

libs/store/src/lib/store.tsx
```diff
@@ -308,10 +308,18 @@ const persistedFcmReducer = persistReducer(
...
+		storage
+	},
+	accountReducer
+);
+
 const reducer = {
 	app: persistedAppReducer,
 	dashboard: dashboardReducer,
-	account: accountReducer,
+	account: persistedAccountReducer,
```
huylexuan44, line 322: Persisting account in the main chat store looks unrelated to admin — confirm no chat/desktop persist regression.

apps/admin/src/app/loader/authLoader.ts
sadohpv: isLogin is in authReducer. AccountReducer only have data user not stand for isLogin

libs/store/src/lib/store.tsx
sadohpv: chat persist for not call BE on admin to get only name and avatar

Issue #13241, opened 2026-06-04
[Desktop/Website] Lost typing indicator in Topic


sheisThanhHien:
https://github.com/user-attachments/assets/fcefaf99-40b1-447c-b54d-30ddcd8a33a5

Luc1404:
<img width="1903" height="944" alt="Image" src="https://github.com/user-attachments/assets/c5030431-1b3d-4592-8779-d8d0b43dc750" />

sheisThanhHien:
## 15062026 - Failed on dev

**Case:** Vẫn mất typing indicator lúc ban đầu mở topic
**Steps to reproduce:**
1. User A create topic
2. User B open and chat on topic
3. Both user refresh app
4.  User A open topic --> User B open and chat on it --> User A cannot see User B typing

**Evidence:**

https://github.com/user-attachments/assets/2bb2bade-96ec-43bd-9095-32a39306fdfd

PR #13306, opened 2026-06-16
260616 - WEB/DESKTOP - lost typing refesh app

Luc1404:
Desktop/Website] Lost typing indicator in Topic
https://github.com/mezonai/mezon/issues/13241

### DIFF
--- a/apps/chat/src/app/layouts/ClanLayout.tsx
+++ b/apps/chat/src/app/layouts/ClanLayout.tsx
@@ -11,6 +11,7 @@ import {
 	selectCurrentClanId,
 	selectCurrentClanIsOnboarding,
 	selectCurrentClanName,
+	selectCurrentTopicId,
 	selectIsShowChatStream,
 	selectIsShowCreateThread,
 	selectIsShowCreateTopic,
@@ -91,6 +92,7 @@ const ClanLayout = () => {
 	const currentChannelType = useSelector(selectCurrentChannelType);
 	const isShowCreateThread = useSelector((state) => selectIsShowCreateThread(state, currentChannelId as string));
 	const isShowCreateTopic = useSelector(selectIsShowCreateTopic);
+	const currentTopicId = useSelector(selectCurrentTopicId);
 	const chatStreamRef = useRef<HTMLDivElement | null>(null);
 	const dispatch = useAppDispatch();
 	const { setSubPanelActive } = useGifsStickersEmoji();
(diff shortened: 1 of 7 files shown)


apps/chat/src/app/layouts/ClanLayout.tsx
```diff
@@ -114,6 +116,13 @@ const ClanLayout = () => {
 		}
 	}, [currentClanIsOnboarding, currentClanId, dispatch]);
 
+	useEffect(() => {
+		if (isShowCreateTopic && currentTopicId) {
```
huylexuan44, line 120: Effect sets focusTopicBox true on open but never false on close — add else branch or stale focus breaks reactions/emoji paths.

libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -1472,13 +1473,13 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
 			if (!isDM) {
 				const currentChannelId = selectCurrentChannelId(state as unknown as RootState);
 				const currentTopicId = selectCurrentTopicId(state as unknown as RootState);
-				const isFocusTopicBox = selectClickedOnTopicStatus(state as unknown as RootState);
+				const isShowCreateTopic = selectIsShowCreateTopic(state as unknown as RootState);
 
 				if (isTopicTyping) {
-					if (!isFocusTopicBox || !currentTopicId || channelId !== currentTopicId) return;
+					if (!isShowCreateTopic || !currentTopicId || channelId !== currentTopicId) return;
```
huylexuan44, line 1480: Typing now gates on isShowCreateTopic but onmessagereaction still uses isFocusTopicBox — topic reactions may stay broken after refresh.

libs/core/src/lib/chat/contexts/ChatContext.tsx
```diff
@@ -1472,13 +1473,13 @@ const ChatContextProvider: React.FC<ChatContextProviderProps> = ({ children, isM
...
 					if (e.channel_id !== currentChannelId) return;
 				} else {
-					if (isFocusTopicBox && currentTopicId) return;
+					if (isShowCreateTopic && currentTopicId) return;
```
huylexuan44: Channel typing suppressed while isShowCreateTopic even after topbar clears focus — old path used isFocusTopicBox; confirm intended.

apps/chat/src/app/pages/topicDiscussion/index.tsx
```diff
@@ -19,6 +19,7 @@ const TopicDiscussionMain = () => {
 	useEscapeKey(() => {
 		setIsShowCreateTopic(false);
 		dispatch(topicsActions.setCurrentTopicId(''));
+		dispatch(topicsActions.setFocusTopicBox(false));
```
huylexuan44, line 22: setFocusTopicBox(false) only on Escape and TopicHeader — channelLoader/clanLoader/MessageContextMenu close paths still omit it.

libs/core/src/lib/chat/contexts/ChatContext.tsx
Luc1404: suppressing channel typing on isShowCreateTopic was not intended. Restored isFocusTopicBox for the channel-typing else branch; kept isShowCreateTopic only for topic-typing ingress

Issue #13309, opened 2026-06-16
[Backend] User info not updated in DM list after changing avatar and display name


sheisThanhHien:
<img width="643" height="1065" alt="Image" src="https://github.com/user-attachments/assets/5113009f-6738-4b17-a35e-9eccf37324ef" />

PR #13326, opened 2026-06-17
260617 -  WEB/DESKTOP - User info not updated in DM list after changing avatar and display name

Luc1404:
[Backend] User info not updated in DM list after changing avatar and display name
https://github.com/mezonai/mezon/issues/13309


libs/components/src/lib/components/DmList/DMListItem/index.tsx
```diff
@@ -125,9 +125,7 @@ function DMListItem({ id, currentDmGroupId, joinToChatAndNavigate, navigateToFri
 	);
 }
 
-export default memo(DMListItem, (prev, cur) => {
-	return prev.id === cur.id && prev.isActive === cur.isActive;
-});
+export default DMListItem;
```
huylexuan44: Custom memo blocked updates when only Redux changed; prefer memo(DMListItem) default compare instead of removing memo.

libs/store/src/lib/direct/direct.slice.ts
```diff
@@ -772,6 +795,38 @@ export const directSlice = createSlice({
...
+		updateCurrentUserInDMs: (state, action: PayloadAction<{ userId: string; displayName?: string; avatarUrl?: string }>) => {
+			const { userId, displayName, avatarUrl } = action.payload;
+			if (!userId) return;
+
+			Object.keys(state.entities).forEach((dmId) => {
+				const dmGroup = state.entities[dmId];
+				if (!dmGroup?.user_ids?.includes(userId) || dmGroup.type !== ChannelType.CHANNEL_TYPE_DM) return;
+
+				const isSelfDm = dmGroup.user_ids.length > 0 && dmGroup.user_ids.every((id) => id === userId);
+				if (!isSelfDm) return;
```
huylexuan44, line 794: updateCurrentUserInDMs only updates self-DMs; confirm #13309 is self-DM — 1:1 rows show the other user.

libs/store/src/lib/direct/direct.slice.ts
```diff
@@ -772,6 +795,38 @@ export const directSlice = createSlice({
 				dmGroup.display_names[index] = display_name;
 			}
 		},
+		updateCurrentUserInDMs: (state, action: PayloadAction<{ userId: string; displayName?: string; avatarUrl?: string }>) => {
+			const { userId, displayName, avatarUrl } = action.payload;
+			if (!userId) return;
+
+			Object.keys(state.entities).forEach((dmId) => {
```
huylexuan44, line 789: Mutates entities in place — use directAdapter.updateOne per dm so selectors and persist see new entity refs.

libs/components/src/lib/components/ModalUserProfile/index.tsx
```diff
@@ -200,23 +200,31 @@ const ModalUserProfile = ({
 	const handleOnKeyPress = useCallback(
 		(e: React.KeyboardEvent<HTMLInputElement>) => {
 			if (e.key === 'Enter' && content && onLoading.current === false) {
+				if (isFooterProfile && userProfile?.user) {
```
huylexuan44: ModalUserProfile Enter-key changes look unrelated to DM list sync — split or note in the PR description.

libs/store/src/lib/direct/direct.slice.ts
```diff
@@ -763,6 +779,13 @@ export const directSlice = createSlice({
 
 			if (avatar && dmGroup.channel_avatar) dmGroup.channel_avatar = avatar;
```
huylexuan44, line 767: channel_avatar still guarded by truthy check; 1:1 rows use avatars[] — OK, but group rows may skip avatar sync.

libs/components/src/lib/components/SettingProfile/SettingRightUserProfile/index.tsx
```diff
@@ -93,6 +94,13 @@ const SettingRightUser = ({
 						avatarUrl: urlImage
 					})
 				);
+				dispatch(
+					directActions.updateCurrentUserInDMs({
```
huylexuan44, line 94: Dispatches updateCurrentUserInDMs after profile save — good; ensure urlImage is the post-upload URL, not a local blob.

Luc1404: urlImage is only set from attachment.url after handleUploadFile / processImage, or from the existing server avatar prop. The blob: URL is used only for ImageEditor preview (imageObject.url), never stored in urlImage

libs/store/src/lib/direct/direct.slice.ts
Luc1404: #13309 is self-DM only. Regular 1:1 rows display the other user (avatars.at(-1), channel_label = peer name); addGroupUserWS excludes current user from user_ids for 1:1 DMs. isSelfDm intentionally limits updates to rows where all user_ids are the current user.

Issue #13225, opened 2026-06-02
[Desktop/Website] Notification settings bug


sheisThanhHien:
### 1. UI bugs: Sai placeholder text
**Actual Result:** `Select or create an option ... `
**Expected Result:** `Select a channel or category...` (apply i18n) 

<img width="734" height="537" alt="Image" src="https://github.com/user-attachments/assets/70f7d652-453e-41d9-a481-ed9cfbfb450b" />

### 2. Sai behavior khi search/chọn channel hoặc category
**Actual Result:**
- Khi chọn 1 channel, channel vẫn nằm ở placeholder 
- Khi search 1 category/channel/thread ko có, hiện Create channel đó --> SAI

https://github.com/user-attachments/assets/9eb88bda-481f-4d50-baf5-435446afb46e

**Expected Result:** 
- Khi search 1 category/channel/thread  không có --> Hiện `No results found.`

https://github.com/user-attachments/assets/0c1ec57f-98fa-4cce-bf53-b564c94d8dc7

### 3. Setting từ phía channel xong mở bảng setting all nó lỗi hiển thị với lấy luôn cái state của channel vừa set làm default

<img width="900" height="534" alt="Image" src="https://github.com/user-attachments/assets/9dbe60bb-fdd2-429c-a11b-80b181445e4e" />

sheisThanhHien:
## 18062026 - Failed on dev

Bug: Select channel/category/thread --> default là Unmute --> F5 --> Auto mark là mute hết

https://github.com/user-attachments/assets/f2f92da9-4f74-4aaf-ba4a-5b4ad0c968e0

PR #13337, opened 2026-06-19
260819 - WEB/DESKTOP - Notification settings bug mute channel

Luc1404:
[Desktop/Website] Notification settings bug
https://github.com/mezonai/mezon/issues/13225

### DIFF
--- a/libs/components/src/lib/components/NotificationSetting/index.tsx
+++ b/libs/components/src/lib/components/NotificationSetting/index.tsx
@@ -119,7 +119,7 @@ const ModalNotificationSetting = (props: ModalParam) => {
 				return isMutedFromMuteTime(muteTime);
 			}
 
-			return false;
+			return (setting.action ?? 1) === 0;
 		},
 		[categoryNotificationSettings, channelNotificationMuteSeconds]
 	);
--- a/libs/components/src/lib/components/NotificationSetting/index.tsx
+++ b/libs/components/src/lib/components/NotificationSetting/index.tsx
@@ -112,16 +112,14 @@ const ModalNotificationSetting = (props: ModalParam) => {
 	};
 	const getIsMuted = useCallback(
 		(setting: (typeof channelCategorySettings)[number]) => {
-			const muteTime =
-				setting.channel_category_title === 'category'
-					? categoryNotificationSettings[setting.id]?.time_mute
-					: channelNotificationMuteSeconds[setting.id];
+			const isCategory = setting.channel_category_title === 'category';
+			const muteTime = isCategory ? categoryNotificationSettings[setting.id]?.time_mute : channelNotificationMuteSeconds[setting.id];
 
-			if (muteTime !== undefined && muteTime !== null && muteTime !== EMuteState.UN_MUTE) {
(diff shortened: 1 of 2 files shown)


libs/components/src/lib/components/NotificationSetting/index.tsx
```diff
@@ -112,16 +112,14 @@ const ModalNotificationSetting = (props: ModalParam) => {
...
+			const isCategory = setting.channel_category_title === 'category';
+			const muteTime = isCategory ? categoryNotificationSettings[setting.id]?.time_mute : channelNotificationMuteSeconds[setting.id];
 
-			if (muteTime !== undefined && muteTime !== null && muteTime !== EMuteState.UN_MUTE) {
+			if (muteTime !== undefined && muteTime !== null) {
 				return isMutedFromMuteTime(muteTime);
 			}
 
-			return (setting.action ?? 1) === 0;
+			return false;
```
huylexuan44: return false drops setting.action fallback; muted rows may show unmuted until per-item fetches finish.

libs/store/src/lib/notificationSetting/notificationSettingCategory.slice.ts
```diff
@@ -423,11 +424,26 @@ export const fetchChannelCategorySetting = createAsyncThunk(
 				};
 			}
 
+			const mappedList = response.notification_channel_category_settings_list.map(mapChannelCategorySettingToEntity);
+
+			await Promise.all(
```
huylexuan44: Promise.all fires one getNotificationSetting/getDefaultNotificationCategory per list row — N network calls on each open.

libs/store/src/lib/notificationSetting/notificationSettingCategory.slice.ts
```diff
@@ -423,11 +424,26 @@ export const fetchChannelCategorySetting = createAsyncThunk(
...
+				mappedList.map((item) => {
+					if (!item.id) {
+						return Promise.resolve();
+					}
+
+					if (item.channel_category_title === 'category') {
+						return thunkAPI.dispatch(getDefaultNotificationCategory({ categoryId: item.id, clanId, noCache: true }));
```
huylexuan44: noCache: true on every item bypasses cached mute state; consider noCache: false after first load.

libs/store/src/lib/notificationSetting/notificationSettingCategory.slice.ts
```diff
@@ -423,11 +424,26 @@ export const fetchChannelCategorySetting = createAsyncThunk(
...
+					}
+
+					return thunkAPI.dispatch(getNotificationSetting({ channelId: item.id, noCache: true }));
+				})
+			);
```
huylexuan44, line 442: If any child dispatch rejects, Promise.all fails and the category list fetch is rejected too.

huylexuan44:
Review notes:
- getIsMuted now ignores setting.action; UI may flash wrong mute state until child fetches complete.
- fetchChannelCategorySetting fans out N parallel API calls with noCache: true on every list load.
- Promise.all ties list success to every child fetch; one failure rejects the whole thunk.

PR #13490, opened 2026-07-17
fix(chat): defer reconnect sync to avoid socket disconnect loop

huylexuan44:
## Summary

- After socket reconnect, run `reconnectSync` (minimal: active channel messages + staggered `joinClan`) before scheduling full `refreshApp`.
- Full refresh uses `skipReconnectWarmup: true` to avoid duplicate WS joins and channel refetch.
- Stagger `joinClan` by 400ms to reduce server write-queue pressure (WQ_STALL ~5s disconnect loop).
- Cancel pending reconnect timers on remount / repeated reconnect events.

## Root cause

Server closes connections when the TLS write queue stays full for 5s ("slow client"). Immediate post-reconnect `refreshApp` burst (`joinClan` × N) caused sustained backpressure.

## Test plan

- [ ] Open a channel, toggle network offline/online → socket stable, no ~5s disconnect loop
- [ ] Messages in current channel appear after reconnect (~750ms)
- [ ] Sidebar / DM list refresh after ~3.75s
- [ ] Rapid reconnect twice within 2s → no stacked duplicate sync
- [ ] Rate-limited `refreshApp` still runs minimal sync when warmup not skipped

### DIFF
--- a/apps/chat/src/app/layouts/MainLayout.tsx
+++ b/apps/chat/src/app/layouts/MainLayout.tsx
@@ -23,11 +23,14 @@ import { selectTotalUnreadDM, useAppSelector } from '@mezon/store';
 import { MezonSuspense } from '@mezon/transport';
 import { SubPanelName } from '@mezon/utils';
 
-import { memo, useContext, useEffect, useMemo } from 'react';
+import { memo, useCallback, useContext, useEffect, useMemo, useRef } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { Outlet } from 'react-router-dom';
 import ChannelVoice from '../pages/channel/ChannelVoice';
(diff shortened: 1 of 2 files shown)


huybuidoanquang (approved):

huylexuan44:
Opened verification revert PR: #13494

Temporarily restoring the pre-#13490 reconnect flow (`refreshApp()` immediately + in-thunk `fetchMessages`) to confirm whether missing/stuck realtime messages are caused by the new reconnect timing or by existing `addNewMessage` / viewport logic.

If the message bug still reproduces on #13494, root cause is likely in message store/viewport, not this reconnect change. If it disappears, we should keep the stagger/delay from #13490 and land a follow-up fix for message persistence + viewport heal.

PR #13506, opened 2026-07-23
260723 - WEB - fix send flick message

sadohpv:
### Changes
- Move add message from user who send it after response from ack 
- When socket event emit data can pass if it already have 

### Test 
- Send message in channel 
- Send message in DM 
- Spam send message


libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -1808,6 +1820,12 @@ export const messagesSlice = createSlice({
 				case TypeMessage.Location:
 				case TypeMessage.Poll:
 				case TypeMessage.Chat: {
+					if (isMe) {
+						const existMessage = state.channelMessages[channelId].entities[messageId];
+						if (existMessage.id === messageId) {
```
huylexuan44: existMessage can be undefined (echo arrives before ack, or from another device/tab) so undefined.id throws inside the reducer. Optional-chain it; the id compare is also redundant since entities are keyed by id.

libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -1808,6 +1820,12 @@ export const messagesSlice = createSlice({
 				case TypeMessage.Location:
 				case TypeMessage.Poll:
 				case TypeMessage.Chat: {
+					if (isMe) {
+						const existMessage = state.channelMessages[channelId].entities[messageId];
```
huylexuan44: Topic self-messages are stored under topic_id, not channelId, so this dedup misses them and hits the same undefined crash. Resolve the target bucket first.

libs/store/src/lib/messages/messages.slice.ts
```diff
@@ -1134,6 +1134,11 @@ export const editMessageViaApi = createAsyncThunk('messages/editMessageViaApi',
 	}
 });
 
+export const addRealMessage = createAsyncThunk('chat/addRealMessage', async (payload: MessagesEntity, thunkAPI) => {
```
huylexuan44, line 1137: addRealMessage is an async thunk that only dispatches one sync action — dispatch messagesActions.addFakeMessage directly. The name is also misleading (it calls addFakeMessage).

huylexuan44 (approved):

PR #13510, opened 2026-07-24
[WIP] 260724 - WEB - fix default send on public

sadohpv:
### Changes 

- Fix default send on public channel is true

### DIFF
--- a/libs/core/src/lib/policy/context/PermissionContext.tsx
+++ b/libs/core/src/lib/policy/context/PermissionContext.tsx
@@ -1,6 +1,7 @@
-import { getStore, selectCurrentClanId, selectMaxPermissionForChannel, selectUserMaxPermissionLevel } from '@mezon/store';
+import { getStore, selectChannelById, selectCurrentClanId, selectMaxPermissionForChannel, selectUserMaxPermissionLevel } from '@mezon/store';
 import { EOverriddenPermission, EPermission } from '@mezon/utils';
-import React, { ReactNode, createContext, useCallback, useContext, useMemo } from 'react';
+import type { ReactNode } from 'react';
+import React, { createContext, useCallback, useContext, useMemo } from 'react';
 import { useSelector } from 'react-redux';
 import { usePermissionsLevel } from '../hooks/permissions/usePermissionsLevels';
 import { useIsClanOwner } from '../hooks/useIsClanOwner';
@@ -27,7 +28,14 @@ export const PermissionProvider: React.FC<{ children: ReactNode }> = ({ children
 
 				const store = getStore();
 				const overriddenPermissions = selectMaxPermissionForChannel(store.getState(), channelId || '');
-				return overriddenPermissions[permission as unknown as EOverriddenPermission];
+				const targetChannel = selectChannelById(store.getState(), channelId);
+
+				if (permission === EOverriddenPermission.sendMessage && !targetChannel?.channel_private) {
+					return true;
+				}
+				if (targetChannel?.channel_private) {
+					return overriddenPermissions[permission as unknown as EOverriddenPermission];
+				}
 			}
 
 			if (permission === EPermission.clanOwner) {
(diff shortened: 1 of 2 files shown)


libs/core/src/lib/policy/context/PermissionContext.tsx
```diff
@@ -27,7 +28,14 @@ export const PermissionProvider: React.FC<{ children: ReactNode }> = ({ children
 
 				const store = getStore();
 				const overriddenPermissions = selectMaxPermissionForChannel(store.getState(), channelId || '');
-				return overriddenPermissions[permission as unknown as EOverriddenPermission];
+				const targetChannel = selectChannelById(store.getState(), channelId);
+
+				if (permission === EOverriddenPermission.sendMessage && !targetChannel?.channel_private) {
```
huylexuan44: Regression: public channels now ignore a server send-message deny, so read-only channels show a composer and every send gets 403. Default to true only when the override is undefined.

libs/core/src/lib/policy/context/PermissionContext.tsx
```diff
@@ -27,7 +28,14 @@ export const PermissionProvider: React.FC<{ children: ReactNode }> = ({ children
...
+					return true;
+				}
+				if (targetChannel?.channel_private) {
```
huylexuan44: delete-message and manage-thread on public channels now fall through to the clan level check (seeded at 1), so most custom roles get Delete on others' messages.

libs/core/src/lib/policy/context/PermissionContext.tsx
huylexuan44: Server resolves thread permissions from the parent channel; using the thread's own channel_private treats a public thread under a private channel as public.

libs/core/src/lib/policy/context/PermissionContext.tsx
```diff
@@ -27,7 +28,14 @@ export const PermissionProvider: React.FC<{ children: ReactNode }> = ({ children
 
 				const store = getStore();
 				const overriddenPermissions = selectMaxPermissionForChannel(store.getState(), channelId || '');
-				return overriddenPermissions[permission as unknown as EOverriddenPermission];
+				const targetChannel = selectChannelById(store.getState(), channelId);
```
huylexuan44: selectChannelById ignores clanId and returns null for unloaded, topic or other-clan channels, which are then treated as public. Use selectChannelByIdAndClanId.

libs/utils/src/lib/utils/index.ts
```diff
@@ -150,7 +150,7 @@ export const checkSameDay = (startTimeString: string, endTimeString: string) =>
 export const uniqueUsers = (mentions: IMentionOnMessage[], memUserIds: string[] | null, rolesClan: IRolesClan[], refereceSenderId: string[]) => {
 	const uniqueUserId1s = Array.from(
 		new Set(
-			mentions.reduce<string[]>((acc, mention) => {
+			mentions?.reduce<string[]>((acc, mention) => {
```
huylexuan44: mentions.some inside rolesClan.reduce still dereferences undefined mentions, so the crash only moves down. Default mentions to [].

libs/core/src/lib/policy/context/PermissionContext.tsx
```diff
@@ -27,7 +28,14 @@ export const PermissionProvider: React.FC<{ children: ReactNode }> = ({ children
...
+
+				if (permission === EOverriddenPermission.sendMessage && !targetChannel?.channel_private) {
+					return true;
```
huylexuan44: Ban ingestion (upsertBanFromChannel in channel.members.ts) still gates on the raw override, so a banned user on a public channel now gets a composer.

libs/core/src/lib/policy/context/PermissionContext.tsx
```diff
@@ -27,7 +28,14 @@ export const PermissionProvider: React.FC<{ children: ReactNode }> = ({ children
...
+				}
+				if (targetChannel?.channel_private) {
+					return overriddenPermissions[permission as unknown as EOverriddenPermission];
```
huylexuan44: overriddenPermissions[...] is undefined until the fetch resolves and callers only block on === false, so private channels allow send while loading.

libs/core/src/lib/policy/context/PermissionContext.tsx
```diff
@@ -27,7 +28,14 @@ export const PermissionProvider: React.FC<{ children: ReactNode }> = ({ children
 
 				const store = getStore();
```
huylexuan44: channel_private is read via getStore and not subscribed, so a public/private toggle leaves MessageBox and menus on the stale result.

PR #13534, opened 2026-08-03
260803 - WEB - remove retry invalid session_id

sadohpv:
### Changes 

- Remove try connect when not network error 

### Test 

- Connect normal 
- Connect wrong id 
- Connect with offline network

### DIFF
--- a/libs/store/src/lib/BootstrapGate.tsx
+++ b/libs/store/src/lib/BootstrapGate.tsx
@@ -96,7 +96,7 @@ export function BootstrapGate({ children, persistor, fallback }: Props) {
 						}
 
 						try {
-							const result = await connectSocket();
+							await connectSocket();
 							connectOk = true;
 							break;
 						} catch (error) {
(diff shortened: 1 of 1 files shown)


libs/store/src/lib/BootstrapGate.tsx
```diff
@@ -100,12 +100,8 @@ export function BootstrapGate({ children, persistor, fallback }: Props) {
...
 							break;
 						} catch (error) {
-							if (i === MAX_RETRIES) break;
-							const baseDelay = INITIAL_DELAY * Math.pow(2, i);
-							// Add jitter time for not all client call in same time
-							const nextDelay = baseDelay + Math.random() * 500;
-							console.error(`Connection failed. Retrying attempt ${i} in ${nextDelay}ms...`);
-							await delay(nextDelay);
+							console.error('ERROR_CONNECT', error);
+							break;
```
huylexuan44, line 104: Breaking on every connectSocket error logs the user out on the first transient failure — only bail for invalid session; keep retries for network/transient errors.

sadohpv: Retries for network is on top of this case

PR #13580, opened 2026-08-13
260813 - WEB - Fix responsive problem on mobile web

kmbigdragon:
## What's happening
-There's no back to menu button on mobile web's chat UI.
-States of opening profile and list member is not reset to false on opening a channel/thread/DM that leads to immediate jump to profile or list member tab.
-Channel welcome/onboarding block on mobile app is overflowing due to hardcoded w-[400px].

## Fixes
Fixed the problems listed above.

## Tests
This video contains tests for all above bugs.

https://github.com/user-attachments/assets/35932a63-631c-4c13-8908-05f37e8aa533

### DIFF
--- a/libs/components/src/lib/components/ChannelTopbar/index.tsx
+++ b/libs/components/src/lib/components/ChannelTopbar/index.tsx
@@ -114,7 +114,7 @@ const ChannelTopbar = memo(() => {
 	return (
 		<div
 			onMouseDown={onMouseDownTopbar}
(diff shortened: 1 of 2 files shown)


libs/components/src/lib/components/ChannelTopbar/index.tsx
sadohpv, line 147: Check if this function is used

kmbigdragon: Merged handleOpenMenu into openMenu for cleaner code and better understanding for devs

PR #13719, opened 2026-09-11
260910-web : Fix UI in toppic when send first message

kind6688:
https://mello.mezon.vn/boards/ab1d77c0-6b09-4935-8aac-0441b8b38160/tickets/MEZON-455


huybuidoanquang:
@huylexuan44 fixed

PR #13772, opened 2026-09-17
260917-Web- Custom status duration

kind6688:


libs/components/src/lib/components/ModalUserProfile/StatusProfile/ModalCustomStatus.tsx
```diff
@@ -69,7 +85,7 @@ const ModalCustomStatus = ({ name, status, onClose, time_reset = 0 }: ModalCusto
 	const currentClanId = useSelector(selectCurrentClanId);
 
 	const handleSaveCustomStatus = () => {
-		const trimmedStatus = customStatus.trim();
+		const trimmedStatus = customStatus.trim().replace(/\\/g, '\\\\').replace(/"/g, '\\"').replace(/'/g, "''");
```
huylexuan44: Regression: escaping on save re-escapes the already-escaped stored status, so it''s becomes it''''s on each save. Backend query is parameterized; drop the escaping.

libs/components/src/lib/components/ModalUserProfile/StatusProfile/ModalCustomStatus.tsx
```diff
@@ -12,47 +12,63 @@ type ModalCustomStatusProps = {
...
 
+	useEffect(() => {
+		if (!isDirtyRef.current) {
+			setCustomStatus(status ?? '');
+		}
+	}, [status]);
+
+	useEffect(() => {
+		setTimeSetReset(getTimeResetLabel(time_reset, Boolean(noClear), t));
+		setResetTimerStatus(noClear ? 0 : time_reset);
```
huylexuan44: For Today, time_reset is the minutes computed at save time; reopening reuses that stale count, so an evening edit clears the status next day. Keep 0 for Today.

libs/components/src/lib/components/FooterProfile/index.tsx
```diff
@@ -275,9 +275,10 @@ function FooterProfile({ name, status, avatar, userId, isDM, username }: FooterP
 				name={name}
 				onClose={handleCloseModalCustomStatus}
 				time_reset={userMemberStatus?.time_reset}
+				noClear={userMemberStatus?.no_clear}
```
huylexuan44: no_clear/time_reset live only in Redux and no API returns them, so after reload the modal defaults to Today and a Don't clear status is wiped at midnight.

libs/components/src/lib/components/ModalUserProfile/StatusProfile/ModalCustomStatus.tsx
```diff
@@ -12,47 +12,63 @@ type ModalCustomStatusProps = {
...
+	const [noClearStatus, setNoClearStatus] = useState<boolean>(Boolean(noClear));
 	const [customStatus, setCustomStatus] = useState<string>(status ?? '');
 
+	useEffect(() => {
+		if (!isDirtyRef.current) {
+			setCustomStatus(status ?? '');
+		}
+	}, [status]);
+
+	useEffect(() => {
```
huylexuan44: No dirty guard: a CustomStatusEvent or language switch while the modal is open overwrites the duration the user just picked.

libs/store/src/lib/channelmembers/channel.members.ts
```diff
@@ -309,7 +309,8 @@ export const updateCustomStatus = createAsyncThunk(
 				return {
 					status: customStatus,
 					user_id: userId || '',
-					time_reset: minutes
+					time_reset: minutes,
+					no_clear: noClear
```
huylexuan44: Clearing status sends noClear: true, which is now stored and makes the next open default to Don't clear. Don't persist no_clear from a clear.

libs/components/src/lib/components/ModalUserProfile/StatusProfile/ModalCustomStatus.tsx
```diff
@@ -12,47 +12,63 @@ type ModalCustomStatusProps = {
...
+
 	const setStatusTimer = useCallback(
 		(minutes: number, noClear: boolean, option: string) => {
 			setTimeSetReset(option);
 			if (noClear) {
 				setNoClearStatus(true);
+				setResetTimerStatus(0);
 			} else {
 				setNoClearStatus(false);
 				if (option === t('timeOptions.today')) {
```
huylexuan44: End-of-day minutes are computed here at click time and again in the thunk; floor gives 0 in the last minute, which means never clear. Compute once at save, use ceil.

libs/components/src/lib/components/ModalUserProfile/StatusProfile/ModalCustomStatus.tsx
```diff
@@ -12,47 +12,63 @@ type ModalCustomStatusProps = {
...
 
 	useEffect(() => {
 		dispatch(userClanProfileActions.setShowModalFooterProfile(false));
 	}, [dispatch]);
 
 	const handleChangeCustomStatus = (e: React.ChangeEvent<HTMLInputElement>) => {
-		const updatedStatus = e.target.value.slice(0, 128).replace(/\\/g, '\\\\').replace(/"/g, '\\"').replace(/'/g, "''");
-		setCustomStatus(updatedStatus);
+		isDirtyRef.current = true;
+		setCustomStatus(e.target.value.slice(0, 128));
```
huylexuan44: Controlled input seeded with the escaped status can exceed 128 chars; slice then silently drops the last chars on the next keystroke.

libs/components/src/lib/components/ModalUserProfile/StatusProfile/ModalCustomStatus.tsx
```diff
@@ -12,47 +12,63 @@ type ModalCustomStatusProps = {
...
 	};
 
-	function getTimeResetLabel(minutes: number, t: (key: string) => string): string {
+	function getTimeResetLabel(minutes: number, isNoClear: boolean, t: (key: string) => string): string {
```
huylexuan44: getTimeResetLabel is used in the effect but missing from its deps and shadows t; move it to module scope.

